PR #352, opened 2012-08-31
Improve collapse/expand using arrow keys

Infocatcher:
1. Allow collapse tabs using arrow keys with extensions.treestyletab.focusMode = 0
2. Expand key select next tab, if current tab doesn't contain child tabs (like Total Commander behavior)

Note: not fully tested!


piroor:

Issue #242, opened 2012-02-11
Tab bar splitter improvements


Infocatcher:
1. Reset tab bar size
We already have reset tab bar size functionality via double click:
<pre>
	resetTabbarSize : function TSTBrowser_resetTabbarSize() 
	{
		if (this.isVertical) {
			this.clearTreePref('tabbar.shrunkenWidth');
			this.clearTreePref('tabbar.width');
		}
		else {
			this.clearTreePref('tabbar.height');
			let tabContainerBox = this.getTabContainerBox(this.mTabBrowser);
			tabContainerBox.removeAttribute('height');
			this._tabStripPlaceHolder.height = tabContainerBox.boxObject.height;
		}
		this.updateFloatingTabbar(this.kTABBAR_UPDATE_BY_RESET);
	},
</pre>
Proposed improvements:
1) handle only left-clicks (not a "bug", but more intuitive)
2) make default values for "tabbar.shrunkenWidth" and "tabbar.width" configurable

2. We can do something good via middle-click:
1) collapse tab bar (clickable splitter area are small)
2) or toggle auto hide
3) or user defined action (one of the above) :)

piroor:
Done by fc38c81.

PR #353, opened 2012-08-31
Add configurable action for middle-click on tab bar splitter

Infocatcher:
Second part of #242


piroor:

Issue #403, opened 2012-11-13
Pinned tabs are too small


FrzMe:
Using tree style tabs 0.14.2012111201 with Firefox 17 (beta channel - built from http://hg.mozilla.org/releases/mozilla-beta/rev/036da70b61f9 ) my tabs on the right the pinned tabs are displayed slightly too small (too narrow and too short) as seen here: http://i.imgur.com/NiNh0.png (Standard tab style)

piroor:
I close this issue because a new version of TST is released. If the problem still there, please comment here. Thanks.

PR #474, opened 2013-03-02
Make pinned tabs look correctly

frzme:
This attempts to fix issue #403 while I don't like hardcoding pixel values I couldn't get it to work without doing so (I have little to no addon development experience tho)


piroor:

PR #609, opened 2013-11-18
OS X window controls overlap the nav-bar when the tabs aren't on top

mnoorenberghe:
1aa4e0d reduces the overlap by putting them back to their original position. I'm not sure how you want to fix the whole issue but this is a start.


piroor:

PR #759, opened 2014-08-25
Fix nested tabs which were broken by Mozilla bug 1050447.

mikeconley:
The signature of the code that opens new tabs changed, which
made the search / replace monkeypatch code break. This updates
treestyletabs to the new pattern.

Please see [bug 1057616](https://bugzilla.mozilla.org/show_bug.cgi?id=1057616) and [bug 1050447](https://bugzilla.mozilla.org/show_bug.cgi?id=1050447) for details.


piroor:

Issue #839, opened 2015-02-09
Typo in localization: "Always resotre tabs"


Happy-Dude:
While using TreeStyleTabs, I noticed a typo when it asks me to restore tabs: "Always resotre tabs".

This is an easy fix -- just replace the text with the typo corrected. I will submit a pull request and reference this issue.

Happy-Dude:
Fixed as of https://github.com/piroor/treestyletab/commit/04242a8158cf61a93606a2d9b126c3f450f3cc9c

PR #840, opened 2015-02-09
Fixed 'Always resotre tabs' typo in localization

Happy-Dude:
Used find and sed to do a regex search and replace of text.

Corrects issue #839.


piroor:
Oops, thanks!!

Issue #148, opened 2011-08-14
Restoring session results in different tab groupings\order


irvinm:
This is a duplicate of another reported issue that was accidentally closed (https://github.com/piroor/treestyletab/issues/93)  that I am seeing as well.  For me, it is having the tabs in a vertical (left) orientation with several "groups" of tabs.  Using "Session Manager" to store\restore my sessions, I always get the proper number of tabs but the groups can be moved around and\or placed within another group.

I did a quick example and took screenshots (below).  The "before" shot has 19 high level tabs ... some individual, some groups.  (Check the numbers next to the "groups" to see how many total tabs in that group)  I simply closed Firefox, restarted it, and took another screen shot.  (I did "close" 2 of the groups as they were expanded when restarting ... maybe part of the problem?)   As you can see in the "after" screenshot, there are only 15 groups now but the same number of overall tabs.

http://img233.imageshack.us/img233/685/beforet.jpg
http://img405.imageshack.us/img405/3405/afterbw.jpg

I know this is a more difficult one to reproduce, so I will see if I can do it with fewer tabs and get a reproducible procedure.

I am currently using:  0.12.2011061701.

PR #843, opened 2015-02-20
Don't show selected tab above other browser elements

### LINKED ISSUE #586: Vimperator conflict: Selected tab shows above input window
Conflict with the vimperator addon if the list of tabs go below where the vimperator bar comes up to.

Reproduction:
1. open tabs so that the vimperator input window covers it
2. select a tab near the bottom.
3. Pop up the vimperator input window

The selected tab shows up above the vimperator window. See bottom left of screenshot.
screen shot 2013-11-05 at 1 42 56 pm

timss:
With certain skins/themes Tree Style tab has a tendency to be shown above other parts of Firefox when it should not:

z-index error

(Thanks to @mjessome for screenshot).

One example is Vimperator. For more information see issue [#148@vimperator](https://github.com/vimperator/vimperator-labs/issues/148). In Vimperator's case it will only work as expected when 'mixed' or 'plain' is used due to this code from [`dropshadow.css#50`](https://github.com/piroor/treestyletab/blob/master/skin/classic/treestyletab/square/dropshadow.css):

	/* Transparent tabs are shown above solid tabs.
	We have to set z-index to show all tabs in the same layer. */
	.tabbrowser-tabs[treestyletab-mode="vertical"]
	.tabbrowser-tab,
	.tabbrowser-tabs[treestyletab-mode="vertical"]
	.tabbrowser-arrowscrollbox
	.tabs-newtab-button {
		position: relative;
		z-index: auto;
	}

This patch introduces the same logic to the other base skins (square/metal/sidebar), and aimes to fix #586 in Tree Style Tab. Tested on both Windows and Linux.

Note that I'm not sure if this is the best way to implement such a "fix". I tried to look through the CSS and the logic found in [`themeManager.js`](https://github.com/piroor/treestyletab/blob/master/modules/themeManager.js#L67), but it wasn't obvious how it should be resolved.

### DIFF
--- a/skin/classic/treestyletab/metal/base.css
+++ b/skin/classic/treestyletab/metal/base.css
@@ -379,3 +379,9 @@ tabbrowser[treestyletab-tabbar-position="right"]
 	background: #e4e4e4 !important;
 	border-color: #666666 !important;
 }
+
+/* don't show tabs above other browser elements */
+.tabbrowser-tabs[treestyletab-mode="vertical"]
+  .tabbrowser-tab[selected="true"] {
+	z-index: auto;
+}
--- a/skin/classic/treestyletab/sidebar/sidebar.css
+++ b/skin/classic/treestyletab/sidebar/sidebar.css
@@ -482,3 +482,9 @@ tabbrowser[treestyletab-mode="vertical"]
   .tab-close-button {
 	visibility: visible !important;
 }
+
+/* don't show tabs above other browser elements */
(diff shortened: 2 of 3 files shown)


piroor:
Thanks. Does the commit d6be4e6396fad2a04d52e11125ef55888dc6a470 fix the issue?

timss:
Yep, it works for all skins. Much cleaner solution than mine :+1: 

Thanks for looking into it so quickly, and thanks for maintaining Tree Style Tab.

PR #912, opened 2015-07-09
CSS bug fix for the Sidebar skin

tobia:
By default the CSS for the Sidebar skin gives a left padding of 10px to each tab.

This conflicts with the setting of `extensions.treestyletab.indent.property` = `padding`, which is needed in order to get a native look & feel.

Here is a screenshot that shows the 10px bug. You can see that a new tab (created at level 0) looks almost as if it was a level 1 tab, but not quite (it has the wrong padding of 10px, while level 0 tabs have 0px and level 1 have 12px):

screen shot 2015-07-09 at 21 17 20


piroor:
Is there no side effect for the default configuration? If this is just specific for the case `extensions.treestyletab.indent.property`=`padding`, I'm afraid of new another issue for most users who use TST with default configuration.

piroor:
TST 0.19.x is already obsolete, so I close this.

PR #917, opened 2015-07-23
Repair fullscreen detection. Fixes #903

### LINKED ISSUE #903: Auto hide behaviour is reversed after going into fullscreen once
Using current Nightly (2015-06-14) and TST nightly.

STR:
1. Install TST in a new profile. 
2. Leave auto hide setting as default I.e. Normal window: always show; Full screen: auto hide
3. Enter and exit full screen by a HTML5 youtube video, or use this test case: https://bug1173866.bugzilla.mozilla.org/attachment.cgi?id=8621346

Actual result:
Before entering full screen, tab bar behaves as expected.
After going into full screen, tab bar always shows.
Exiting full screen tab bar auto hides in normal windows.

There are a few changes in Nightly recently about fullscreen that may be relevant:
[Bug 1161802](https://bugzilla.mozilla.org/show_bug.cgi?id=1161802) - Put document in fullscreen state and trigger the event after sizemodechange
[Bug 1173866](https://bugzilla.mozilla.org/show_bug.cgi?id=1173866) - DOM fullscreen broken in e10s (DOM fullscreen fails on the other than the first tab)

jryans:
Original fix mentioned by @heftig in #903.

Converted to PR for easy merging.

### DIFF
--- a/modules/browser.js
+++ b/modules/browser.js
@@ -5123,7 +5123,7 @@ TreeStyleTabBrowser.prototype = inherit(TreeStyleTabWindow.prototype, {
 	onBeforeFullScreenToggle : function TSTBrowser_onBeforeFullScreenToggle() 
 	{
 		if (this.position != 'top') {
-			var isEnteringFullScreenMode = !this.window.fullScreen;
+			var isEnteringFullScreenMode = this.window.fullScreen;
 			// entering to the DOM-fullscreen (ex. YouTube Player)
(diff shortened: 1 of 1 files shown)


piroor:
Hmm... this breaks my Firefox 39.0. I think this should be done only for newer versions of Firefox.

jryans:
Ah, you may be right, sorry... I use Nightly myself, so that's probably the issue.

Issue #910, opened 2015-07-05
Bad interaction with fullscreen mode on Firefox Nightly


HighCommander4:
With Firefox Nightly and Tree Style Tabs installed, the tab bar remains on screen but blank when going into full screen mode (with the full-screen content being sized to fill the remaining space, rather than the whole screen).

When coming back out of full screen mode, the tar bar disappears, making tab navigation basically impossible until the browser is restarted.

Tested with today's nightly on Ubuntu 14.04, with a YouTube video as the full-screen content. Note that the problem occurs with both e10s enabled and e10s disabled.

The problem does not occur with Firefox Release.

ivan:
This happens in Firefox 41.0a2 (2015-07-05) (Aurora / Developer Edition) as well, basically making fullscreen mode unusable.

Daniellynet:
Yep, same bug as in #903.

andrebeat:
I confirm this also started happening to me after I upgraded to Firefox 41 (Delevoper Edition).

ivan:
https://github.com/piroor/treestyletab/commit/5958766997362c151da3f038afbb0ae67c3d7e75 was reverted but it does show how to patch your own copy

itonohito:
Now 41.0b1 in beta stage and have the same issue, especially in youtube (twitch's fullscreen works ok).

piroor:
Does this still happen with the latest development build?
http://piro.sakura.ne.jp/xul/xpi/nightly/

Note: if you cannot install the development build with "unsigned" error, you'll have to disable the signature check. Go to `about:config` and turn `xpinstall.signatures.required` to `false`. The preference won't available on released versions of Firefox 44 and later, so you'll have to use Firefox Developer Edition or Nightly to try the development build of TST.

HighCommander4:
The problem is fixed for me with the latest Firefox Nightly + the latest the Tree Style Tabs development build. Thanks!

andrebeat:
I can confirm that this is also fixed for me (it was fixed when PR #925 was merged). Cheers!

piroor:
Thanks. I close this.

PR #925, opened 2015-08-18
Fix fullscreen issues

upsuper:
I am the Gecko developer who currently works on Fullscreen stuffs. I'm sorry that I didn't realize my change breaks addons like this. The commits fix the issues caused by two changes I made in Gecko. See the description for more details.


upsuper:
This should fix #903 and #910.

piroor:
Thanks a lot!!!

mlissner:
@upsuper, I'm just passing by, but thanks for being so conscientious. Rare is the developer that fixes downstream bugs like this one!

Issue #982, opened 2015-11-09
[Firefox Dev & Nightly] New Bookmark Folder and Properties Broken (and other critical functionality repeatedly broken)


xeonx1:
"New Bookmark Folder" and right click (bookmark/folder > "Properties" are still broken for Tree Style Tabs. This occurs only with Firefox Developer Edition and Firefox Nightly (at least with 64-bit versions of Dev and Nightly, probably 32-bit as well), and can't be reproduced with Firefox stable.

As seen in the below screenshot, these dialogs show up without any fields except Sort by Title checkbox.  This occurs for both the latest XPI release (v0.16.2015110801) and the latest from GitHub. 

![broken new bookmark folder and properties dialog for firefox dev and nightly](https://cloud.githubusercontent.com/assets/4053734/11027579/91a162d2-866b-11e5-96a6-2f1f6537cd6d.png)

Though the Add Bookmark (star dialog), at least, was finally fixed (for now) for Tree Style Tabs - after being broken continually for a year - the New Folder and Bookmark Properties dialog issues remain.

Even worse, a few days ago, after a Firefox and/or extension update, both Tree Style Tabs (latest XPI release) and Multiple Tab Handler broke critical functionality for Firefox Dev and Nightly 64-bit, resulting in Firefox no longer showing dialogs and drop-downs, including the Firefox main app (hamburger) menu, LastPass, Add Bookmark, New Bookmark folder, and even addon install dialogs. 

Though the last commit/release resolved some of that for Tree Style Tabs, other related issues remain for Tree Style Tabs, and even the latest in GitHub hasn't fixed the more critical issues for Multiple Tab Handler. See the related issue reported for Multiple Tab Handler (https://github.com/piroor/multipletab/issues/101).
____________________

The most recent temporary workaround for this in Tree Style Tab (and similarly in Multiple Tab Handler) is:
   Add "return;" after line 25 in \treestyletab\res\bookmarkMultipleTabs_bookmarkPropertiesOverlay.xul
____________________

[...]

PR #989, opened 2015-11-12
Fix for piroor/treestyletab#982 New Folder and Bookmark Properties

xeonx1:
Fix for piroor/treestyletab#982 and piroor/multipletab#101 as last commit did not fix the issue. Pull requests need to be applied to both repos for fix to work reliable.  Fixes issue where New Folder and Bookmark Properties dialogs appearing empty only both Tree Style Tabs and Multiple Tab Handler (and other extensions withbookmarkMultipleTabs_bookmarkPropertiesOverlay.xul) are enabled at the same time.


piroor:
Oops, thanks!

PR #1021, opened 2016-01-11
Don't force visibility of navigation in fullscreen

regellosigkeitsaxiom:
Do not override user settings on visibility of navigation bar in fullscreen mode.


regellosigkeitsaxiom:
Need further testing.

piroor:
The code is written to move closebox (and other window controls) from the hidden title bar (it is invisible when the tab bar is vertical) to the navigation bar. I think it won't affect to the visibility of the navigation bar itself...

PR #1069, opened 2016-02-16
Repair tearing off a tree of multiple tabs

jryans:
With Firefox 47 and this add-on built from f20f936a030c73b7af0981a2ce7ad2e19d654e03, the context menu item "Move to New Window" would fail for me. A new window would open, but it would only contain a blank new tab, and none of the tabs would move.

This appears to fix this issue in 47 (Nightly) for me. I also tested in 44 (Release), and it works there as well.


piroor:
Thanks!

Issue #940, opened 2015-09-21
More accessible settings/preferences?


SRGOM:
Perhaps in the tab bar context menu itself? I find myself using the settings whenever my tab bar refuses to resize.
Thanks for the great add-on btw.

piroor:
The configuration UI was removed in an old version, and I dithered to re-introduce it because 1) there are already too many configuration items, and 2) it was hard to work with single configuration UI and multiple existing windows. However, I knew that such a configuration UI will help some people to fix the size of unexpectedly expanded tab bar.

By recent commits, now the configuration UI is re-introduced and it always resets the size of the tab bar on all windows. You still can resize the tab bar for each window. You can try a development build: http://piro.sakura.ne.jp/xul/xpi/nightly/

SRGOM:
You're right, I mean just having a link in the context menu to the settings box, rather than one going to add ons -> TST -> preferences

piroor:
Oops, I misunderstood...

I clearly dislike custom menuitem to go to the configuration dialog of the addon itself, because it is hard to be maintained for future versions of Firefox, and too many "configurations..." menutiems provided by various addons are ugly.

Instead, if a configuration item is need to be accessed too frequently, then the configuration should be controlled by a seamless operation like simple drag and drop of the separator between the tab bar and the content area, without any checkbox or textbox.

Which is the configuration item you want to access very frequently to?

SRGOM:
Well, I hav

piroor:
> Well, I have this problem where I am unable to unhide the sidebar after hiding it. So I go to config to change its' position to top first and then to left.

Did you mean the issue?: when you activate the "hide tab bar automatically" feature, sometimes (or always) the hidden tab bar does not reappear. Then, I think it must be tracked as an issue of the auto-hide-tab-bar feature itself, because the tab bar must be shown/hidden by moving of your mouse cursor.

SRGOM:
Yes, that is an issue too. The issue I face though is that even manually hiding makes it hard to pull it back, sometimes at all, or sometimes to its complete width. The solution is to go to preferences.

piroor:
> The issue I face though is that even manually hiding makes it hard to pull it back, sometimes at all, or sometimes to its complete width.

You told about clicking of the "grippy" of the splitter itself (or dragging the splitter to the edge of the window to collapse the tab bar)?

piroor:
[...]

PR #1080, opened 2016-02-23
Add a preferences menu item to the context menu

SRGOM:
The context menu that shows up on left clicking the tree-style-tab bar
now has a preferences button to open preferences quickly

### DIFF
--- a/content/treestyletab/treestyletab.xul
+++ b/content/treestyletab/treestyletab.xul
@@ -142,6 +142,11 @@
 			type="checkbox"
 			autocheck="false"
 			oncommand="TreeStyleTabService.toggleFixed(TreeStyleTabService.getTabBrowserFromChild(this));"/>
+		<menuseparator id="context-separator-preferences"/>
+		<menuitem id="context-item-preferences"
+			label="&context.preferences.label;"
+			accesskey="&context.preferences.accesskey;"
+			oncommand="TreeStyleTabService.showPreferences();"/>
 	</menupopup>
 	<tooltip id="treestyletab-full-tree-tooltip"/>
 	<panel id="treestyletab-autohide-content-area-screen"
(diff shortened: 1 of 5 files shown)


SRGOM:
Resolves https://github.com/piroor/treestyletab/issues/940

SRGOM:
![Image]http://i.imgur.com/bGGHQPK.png?1

SRGOM:
Image

SRGOM:
Unfortunately, I have no experience with firefox addon-writing. I hope the changes I've made are okay, but if they don't seem to, please feel free to discard without spending too much time to review them...

piroor:
Sorry but this is really out of focus of TST. If the configuration dialog is frequently required in your daily use, there is something implicit problem which must be solved in another way. Instead, please describe why you need such a fast pass to the configuration dialog. After the actual problem is solved, you won't need such a menu anymore. In other words, adding such a menu can disguise fatal problems which really should be solved. I'm really worrying about it.

piroor:
BTW, the patch itself seems good for the purpose. I don't know why but if you really really need the menu, you'll use your forked version and merge my changes sometimes.

SRGOM:
Your point is right. I did have problems earlier with the adjustment of
width (sometimes I wouldn't be able to bring the bar back from 0 width) but
it no longer exists so I must have been doing something wrong...

Although one major benefit I see of it is for new users to be comfortable
playing around with settings to customize it the way they want. I myself
wished for it when I was starting to use it. Try a setting for a bit, see
if it improves the workflow. If not, change it..

SRGOM:
I went back and thought about it and I think at least two wildly popular extensions: NoScript and uBlock Origin provide very direct access to preferences. Its also helpful for new users to change things fast and get comfortable

pofl:
@piroor Well I just installed TST today, and for me a quick way to the settings would have been very comfortable as I changed the theme about 20 times and I'm still not sure whether I'll like the current one still tomorrow. I also played around with other options until I liked it somewhat.
I see your reasoning about this. However I would nevertheless strongly make a case for adding this option. It's optional after all.

piroor:

BTW, now major options are configurable directly via the context menu on the toolbar button of TST.

PR #1100, opened 2016-03-31
Update readme

Croydon:
The original Vertical Tabs and Vertical Tabs (Simplified) are both not longer maintained.
Vertical Tabs was getting broken with the latest Firefox updates. The Simplified version is at least still working.

Because missing maintaince of all forks of the Vertical Tabs add-on I have started a fork by myself a while ago. 

This pull request would remove the now unuseable Vertical Tabs and add my fork of the add-on to the readme.


piroor:
Thanks!

PR #1110, opened 2016-04-30
fix #394.

### LINKED ISSUE #394: Use 'new Function()' instead of 'eval()'
This change may speed up a bit & increase maintainability.
Because:
* `new Function()` does not hold context. It doesn't look up context variable.
*  The function made with `new Function()` look up variables as static. It does not depend on context.

lv7777:
simply fix...


piroor:
Thanks! By the way, I think that commit-comment including the reason why we should use "new Function" instead of "eval" will be more better ;)

PR #1192, opened 2016-09-13
Metal: Fix sizing of selected pinned tabs

### LINKED ISSUE #1191: Strange style for pinned tabs using Metal theme
## Short description
The pinned tab style, using Metal theme, becomes squashed when clicked.

## Steps to reproduce

 2. Install TST.
 3. Set the TST Metal theme.
 4. Pin a tab.
 5. Click pinned tab.

## Expected result
The tab should be highlighted and remain same size.

## Actual result
The pinned tab gravity and size changes.

Before click:
screen shot 2016-09-12 at 1 21 53 pm

After click:
screen shot 2016-09-12 at 1 22 05 pm

## Environment

 * Platform (OS): OSX 10.11.6
 * Version of Firefox: 48.0.1 and 50.0a2
 * Version (or revision) of Tree Style Tab: 0.18.2016090802 and Nightly "12-Sep-2016 01:12"

talklittle:
Fixes #1191

### DIFF
--- a/skin/classic/treestyletab/metal/tab.css
+++ b/skin/classic/treestyletab/metal/tab.css
@@ -196,23 +196,6 @@
 
 .tabbrowser-tabs[treestyletab-mode="vertical"]
   .tabbrowser-tab:not(:hover)
-  .tab-content[pinned],
-.tabbrowser-tabs[treestyletab-mode="vertical"]
-  .tabbrowser-tab:hover
-  .tab-content[pinned],
-.tabbrowser-tabs[treestyletab-mode="vertical"]
-  .tabbrowser-tab[visuallyselected]:not(:hover)
-  .tab-content[pinned],
-.tabbrowser-tabs[treestyletab-mode="vertical"]
-  .tabbrowser-tab[visuallyselected]:hover
-  .tab-content[pinned] {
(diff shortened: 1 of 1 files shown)


piroor:
Thanks! I have some questions:

 * If possible, could you describe how this change works? I'm afraid of regressions, so I hope to merge this change without such a worrying.
 * Does this change work as expected on three versions: Firefox 45ESR, Firefox 48, and Nightly 51.0a1? If this change breaks something on Firefox 45ESR, I think I have to do some hack to apply such version-specific rules only for newer versions of Firefox.

talklittle:
Works as expected in ESR 45.3.0, 48.0.1, Developer Edition 50.0a2, and Nightly 51.0a1.

Verified by installing Tree Style Tab from Mozilla Addons, and seeing the original bug (squashed pinned tabs when visually selected). Then replacing it with the locally built .xpi containing this PR's contents.

The change works by looking at the `[visuallyselected]` pinned tabs, and not altering the tab content margin/padding, so the pinned tab (containing just a favicon) can retain its normal size. It still gets highlighted when clicked.

talklittle:
@piroor Thank you!

piroor:
Perfect! Thanks again!

PR #1194, opened 2016-09-13
Fix stretched buttons on Metal theme on OS X

talklittle:
Before fix:
screen shot 2016-09-13 at 11 18 08 am

After fix:
screen shot 2016-09-13 at 11 18 19 am

Affects OS X, not Ubuntu.

Verified fix on ESR 45.3.0, 48.0, 50.0a2, 51.0a1


piroor:
Wow! Thanks a lot!!

PR #1222, opened 2016-11-29
Added greek translation

skarmoutsosv:
A greek translation.

Thank you for your valuable addon. I have spend many years trying to organize my tabs (using bookmarks, groups, social sites etc.) and TST looks like my final destination.


piroor:
Thanks a lot!!

Issue #1460, opened 2017-10-11
proposal: Enable ESLint rules


saneyuki:
## Motivation

https://github.com/piroor/treestyletab/pull/1444 introduced ESLint, but it only checking a parse error. ESLint has many rules to prevent anti-patterns. So I think it's better to enable it.

## Detailed Design

ESLint rules has some aspects:

- _Prevent Errors_
- _Sort styles_

Basically, I think we should make it an _error_ that would causes an bug/error/mistake frequently. But I also think that stylistic problems should be kept as warning. They should not prevent your debugging/coding.

I'll open some pull requests by the above thinking.

@piroor よろしくお願いします

piroor:
了解です。
手間はかかってしまいますが、段階的にルールを反映していければと思っています。

piroor:
I close this because outdated.

PR #1461, opened 2017-10-11
Enable eslint's quote rule

saneyuki:
- https://eslint.org/docs/rules/quotes
- relate: https://github.com/piroor/treestyletab/issues/1460
   - I'd like to enable this as warning because this is a stylistic problem.


piroor:
Thanks!

PR #1558, opened 2017-11-17
Single-source the icon, use fill: currentColor

joar:
According to https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json/icons#SVG, the SVG does not need to exist in multiple resolutions, it will be scaled to the proper resolution, however, it still needs multiple entries in the `{"icons": {"24": "...", "48": "..."}}` parts of the manifest

Using `fill="currentColor"` hopefully means that userChrome.css can affect the color of treestyletab's icon.

The .ai file has just been renamed, I cannot update it since I don't have an Illustrator license, but it should probably be updated as well.


piroor:
Does this work for a regular environment with `svg.context-properties.content.enabled`=`false`? It seems to be disabled by default for current release of Firefox.

piroor:
And please keep each size source .ai files because they are little bit different.

webextensions/manifest.json
```diff
@@ -19,30 +19,30 @@
...
   "icons": {
-    "16": "resources/16x16-light.svg",
-    "20": "resources/20x20-light.svg",
-    "24": "resources/24x24-light.svg",
-    "64": "resources/icon64.png",
-    "94": "resources/icon94.png"
+    "16": "resources/icon.svg",
+    "20": "resources/icon.svg",
+    "24": "resources/icon.svg",
+    "64": "resources/icon",
```
trevorhobson: I don't think the 64 and 94 lines should be changed and the files should not be removed as they are still necessary.

webextensions/manifest.json
```diff
@@ -19,30 +19,30 @@
...
+    "94": "resources/icon"
   },
   "background": {
     "page": "background/background.html"
   },
   "browser_action": {
     "default_title": "__MSG_sidebarTitle__",
     "default_icon": {
```
trevorhobson: According to https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json/browser_action#Choosing_icon_sizes 
> You can supply a single icon as an SVG file, and it will be scaled correctly

so this section could just be `"default_icon": "resources/icon.svg",`

webextensions/manifest.json
```diff
@@ -19,30 +19,30 @@
...
-      "24": "resources/24x24-light.svg"
+      "16": "resources/icon.svg",
+      "20": "resources/icon.svg",
+      "24": "resources/icon.svg"
     },
     "browser_style": true
   },
   "sidebar_action": {
     "default_title": "__MSG_sidebarTitle__",
     "default_icon": {
```
trevorhobson: The same as browser_action this section could just be `"default_icon": "resources/icon.svg",`

piroor:
Sorry I'm still negative to unify icons to single one. Small-scaled icon are rendered badly, like:
image

To avoid such a problem, I designed icons for each size matching to screen pixels.
image

This is one of major problems on small-size icons on low-resolution screen. If you use high-resolution screen (like "Retina" at MacBook) you won't see such bad rendering, but I'm still using low-resolution screen. I hope to keep well-designed icons for each size, until I migrate to high-resolution screen.

joar:
@piroor I understand about the icon sizes.

You are also correct about [`-moz-context-properties` only being allowed for `chrome://` and `resource://` URLs with `svg.context-properties.content.enabled=false`](https://developer.mozilla.org/en-US/docs/Web/CSS/-moz-context-properties). Do you know of any reason it's not enabled for  `moz-extension://`, or could it simply have been overlooked? The `#sidebar-icon` element even has the `-moz-context-properties: fill` attribute set in Firefox Developer Edition 58.

joar:
@piroor I have managed to create a responsive SVG icon file that supports context-properties with a fallback color in case context-properties are not allowed:
image
All three versions of the SVG are merged in one SVG file, the SVG file then displays the correct version depending on `@media` queries inside the SVG itself: https://gist.github.com/joar/acd2f788110f5d41ad6553fbec0ea576?short_path=e7bbebe#file-magic-icon-svg

Gist including HTML demo file: https://gist.github.com/joar/acd2f788110f5d41ad6553fbec0ea576

piroor:
Sorry for this large delay. I've read this PR again to reconsider, and I still think that there is less motivation to merge multiple size icons to single file, with some reasons:

1. Unified SVG file is effective when most part is shared in each version. For example, current "32x32-dark.svg" contains:
   
   * Single symbol to define the shape of the icon
   * Multiple fill definitions for different opacities
   
   On the other hand, the magic icon contains multiple shapes for each size and there is no shared part. 
2. Moreover, unified SVG looks effective when the file need to be referred only with single fixed URL.  Path definitions for each size icons are already supported with manifest.json itself.

And, `svg.context-properties.content.enabled` is still `false` by default. The result I got is:
image

> Do you know of any reason it's not enabled for moz-extension://, or could it simply have been overlooked? 

Sadly I don't know the reason. [1377302 - Support fill="context-fill" for mozilla-signed WebExtension icons}(https://bugzilla.mozilla.org/show_bug.cgi?id=1377302) looks related and its original summary was `Support fill="context-fill" for WebExtension icons`, but it was changed to `Support fill="context-fill" for mozilla-signed WebExtension icons` with no explanation.

ariasuni:
There’s a bug report on allowing this use case though CSS `env()` rather than SVG `fill`:
https://bugzilla.mozilla.org/show_bug.cgi?id=1421329

But there’s still an open bug report to allow `'-moz-context-properties'`:
https://bugzilla.mozilla.org/show_bug.cgi?id=1388193

So I guess we have to watch these issues and wait.

PR #1635, opened 2017-11-30
Zhtw translate

lycsjm:
Update translate and fix some typo, also add some newline in en, ja, and zh_TW locale to make them have same style.


piroor:
Thanks a lot!

Issue #1641, opened 2017-12-01
TST 2.3.0 stops respecting desktop colors in Linux


danielkza:
## Short description

TST 2.2 and earlier used to respect the system colors of my system Theme (Arc-Dark) on Fedora 27. Starting with TST 2.3.0, the Plain theme is always light colored instead:

https://i.imgur.com/n417Af9.png

## Steps to reproduce

Compare TST 2.2 and 2.3 in Linux with any dark GTK theme enabled.

## Expected result

Default theme respects theme colors.

## Actual result

Default theme does not respect theme colors.

## Environment

 * Platform (OS): Linux - Fedora 27
 * Version of Firefox: 57.0
 * Version (or revision) of Tree Style Tab: 2.3.0

piroor:
Sorry I didn't know the system color is used on Linux. By recent commits I've introduced a configuration for color scheme and now "System Color" is the default on Linux.

danielkza:
@piroor Thank you very much! I'll test the latest master and tell you the results.

danielkza:
The latest dev build works perfectly. Awesome work, thank you!

PR #1644, opened 2017-12-01
Add 'System Colors' theme

actionless:


piroor:
To fix the issue #1641 I've introduced some changes to apply system colors only on Linux. So I think there is less reason to add new theme for system colors. How do you think? You'll try the fix on the latest development (unsigned) build: https://piro.sakura.ne.jp/xul/xpi/nightly/treestyletab-we.xpi

actionless:
i've tried your build -- system colorscheme is fixed there.
However on dark colorscheme hover tab style is not clearly distinguishable from a normal tab style. What do you think about renaming the theme in PR from 'System Colors' to 'High Contrast'?

Also i've noticed a (probably unrelated) bug what colortheme is being applied only after browser restart.

piroor:
Hmm, ”High Contrast" seems not appropriate name on some environment. On my Windows 10, the background color of the tab bar is similar to the color of tabs.
image

piroor:
> Also i've noticed a (probably unrelated) bug what colortheme is being applied only after browser restart.

Sorry, fdb372e should fix it.

actionless:
> Sorry, fdb372e should fix it.
Thanks! I've rebuilt the extension and now it's fine.

I also tried to make few screenshots with dark and light system themes with "Flat" TST theme and the one from PR:

flat-light  sc-light


flat-dark sc-dark


mb show theme only if platform is Linux?

piroor:
Hmm, interesting. on my environment (Windows 10) `ThreeDShadow` is enough darker than `ThreeDFace`. and `InactiveBorder` is not. Opposite, `ThreeDShadow` is not enough different and `InactiveBorder` is different from `ThreeDFace`  on your environment.

piroor:
How about `ThreeDDarkShadow` in your environment? I think it should enough different from `ThreeDFace`.

actionless:
with all system themes i've tried that color was just black, btw i am using this page as test:

http://www.iangraham.org/books/xhtml1/appd/update-23feb2000.html

actionless:
I also added one more commit which adds the missing tab borders to this theme.

actionless:
thanks for merging, i'll try to support it if it will require changes on browser updates

piroor:
Oops its my mistake, I just thought to try it on  different branch...

piroor:
Currently still I think this theme is not ready to merge,  because I still need more reasons: why it is better than existing one? user-defined styles doesn't solve it? And it also need more fixes: missing throbber on my environment, missing thumbnail, mismathed name (it should be "High Contrast").

piroor:
> And it also need more fixes: missing throbber on my environment, missing thumbnail, mismathed name (it should be "High Contrast").

Could you add more PRs for them?

actionless:
yes, i can prepare it tomorrow.

what do you think about such style for a throbber?

```css
.tab.active .throbber {
  background-color: HighlightText;
  border-radius: 6px;
}
```

piroor:
I think you should define throbber color via css variable, instead of direct declatation. See also:
https://github.com/piroor/treestyletab/blob/master/webextensions/sidebar/styles/metal/metal.css#L4
https://github.com/piroor/treestyletab/blob/master/webextensions/sidebar/styles/sidebar/sidebar.css#L50
https://github.com/piroor/treestyletab/blob/master/webextensions/sidebar/styles/sidebar/sidebar.css#L60

And please don't forget to define extra padding for overlay scrollbar on macOS.
https://github.com/piroor/treestyletab/blob/master/webextensions/sidebar/styles/metal/metal.css#L43

Like this you need to maintain theme for platform you don't (possibly never) use also. This is the largest reason why I'm inaggressive to add new themes...

piroor:
> Hmm, interesting. on my environment (Windows 10) ThreeDShadow is enough darker than ThreeDFace. and InactiveBorder is not. Opposite, ThreeDShadow is not enough different and InactiveBorder is different from ThreeDFace on your environment.

To solve this problem, I propose to unify color of tab bar (background area) to text color, like:

```css
root {
    background: ButtonText;
}
```

In your screenshot `ButtonText` and `InactiveBorder` seems similar. How about this?

actionless:
`ButtonText` seems to be too much contrast, ie on light colorschemes it's usually a black color, on dark light-light-grey (almost white)

piroor:
By 59dc58e537311b4444d984ad3761dcf009b67ab5 I've introduced new element just for background. It is independent from the tab bar, so you can set both system color background and opacity for it. The current color (65% `ButtonFace` + 35% `ButtonText`) is just an example. Could you create PR to change this to fit your environment? If the color have enough high contrast on my environment also, I'll ship it as a part of next release.

PR #1651, opened 2017-12-03
fix(themes: highcontrast): define closebutton size explicitly

actionless:
sorry for spamming you :)
but i've just noticed what in some condition close button box turns from a square into a not-even rectangle, this rule should fix that


piroor:
it's okay, thanks!

PR #1659, opened 2017-12-07
Fix #1658

### LINKED ISSUE #1658: TST 2.x seems to request a favicon twice for each tab
## Short description
1. TST 2.x sometimes requests a favicon twice for each tab when TST sidebar is opened (I found this when I looked at my local web servers' logs).
2. For a webpage that does not have any favicon, TST 2.x requests the webpage itself to obtain its favicon every time TST sidebar is opened.  This causes not only unexpected network traffic but also the inconvenience reproduced by the following steps.

## Steps to reproduce
1. Start Firefox 57 with clean profile.
2. Install TST 2.x.
3. Open a new container tab.
4. Open a webpage that has no favicon and requires an HTTP authentication in the tab created by step 3.
5. Open TST sidebar.

## Expected result
No dialog popup.

## Actual result
A password dialog will pop up unexpectedly.

If a password is entered in the dialog, the login information is remembered in the global context, not in the container.

## Environment

 * Platform (OS): macOS 10.13.1
 * Version of Firefox: 57.0.1
 * Version (or revision) of Tree Style Tab: 2.3.0

## Cause and quick fix

I think this is because `TabFavIconHelper.loadToImage` is called twice for each iteration in `rebuildAll` as follows:
 1. it calls `buildTab` that calls `onTabBuilt` that calls `loadToImage`, and
 2. it calls `updateTab` with `{forceApply:true}` that calls `onTabFaviconUpdated` that calls `loadToImage`.

The latter passes `loadToImage` the URL of the tab as its favicon's URL if its `favIconUrl` is undefined.  This is why the webpage itself is requested when the sidebar is opened.

A quick fix should be to
 1. omit the call of `loadToImage` in `onTabBuilt` (I'd say this should be satisfactory since `onTabBuilt` doesn't seem the timing to fill a tab element with its contents), and
    ```diff
    --- a/webextensions/sidebar/handlers.js
    +++ b/webextensions/sidebar/handlers.js
    @@ -633,10 +633,6 @@ function onTabBuilt(aTab, aInfo) {
       var throbber = favicon.appendChild(document.createElement('span'));
       throbber.classList.add(kTHROBBER);
       aTab.insertBefore(favicon, label);
    -  TabFavIconHelper.loadToImage({
    -    image: faviconImage,
    -    tab:   aTab.apiTab
    -  });
    
       var counter = document.createElement('span');
       counter.classList.add(kCOUNTER);
    ```
 2. avoid calling `onTabFaviconUpdated` if favicon does not exist even when `forceApply` is true.
    ```diff
    --- a/webextensions/common/tree/base.js
    +++ b/webextensions/common/tree/base.js
    @@ -218,8 +218,7 @@ function updateTab(aTab, aNewState, aOptions = {}) {
         window.onTabLabelUpdated && onTabLabelUpdated(aTab);
       }
     
    -  if (aOptions.forceApply ||
    -      'favIconUrl' in aNewState ||
    +  if ('favIconUrl' in aNewState ||
            TabFavIconHelper.maybeImageTab(aNewState)) {
         window.onTabFaviconUpdated &&
           onTabFaviconUpdated(
    ```

uenoB:
This is a fix for #1658.

### DIFF
--- a/webextensions/common/tree/base.js
+++ b/webextensions/common/tree/base.js
@@ -218,8 +218,7 @@ function updateTab(aTab, aNewState, aOptions = {}) {
     window.onTabLabelUpdated && onTabLabelUpdated(aTab);
   }
 
-  if (aOptions.forceApply ||
-      'favIconUrl' in aNewState ||
+  if ('favIconUrl' in aNewState ||
        TabFavIconHelper.maybeImageTab(aNewState)) {
     window.onTabFaviconUpdated &&
       onTabFaviconUpdated(
--- a/webextensions/sidebar/handlers.js
+++ b/webextensions/sidebar/handlers.js
@@ -633,10 +633,6 @@ function onTabBuilt(aTab, aInfo) {
   var throbber = favicon.appendChild(document.createElement('span'));
   throbber.classList.add(kTHROBBER);
   aTab.insertBefore(favicon, label);
-  TabFavIconHelper.loadToImage({
-    image: faviconImage,
-    tab:   aTab.apiTab
-  });
 
   var counter = document.createElement('span');
   counter.classList.add(kCOUNTER);


piroor:
Thanks! I didn't take time to better handling of favicons, so this helps me a lot.

PR #1674, opened 2017-12-12
Korean language support

vwvww:
Made a new folder 'ko-KR' in locale and translated it from japanese to korean.

Additional test needed.
(I translated and applied it by adding it in chrome.manifest and install.rdf , but some features such as preference are showing en-US...)

日本語での対応もできますので、もし日本語のほうで良ければ日本語でコメントお願いします。


piroor:
Thanks a lot, but these files are for the legacy version and you need to add another file for the current version 2.x. Could you send more PR?
https://github.com/piroor/treestyletab/tree/master/webextensions/_locales

PR #1708, opened 2018-01-05
fix(styles: highcontrast): use Background/Caption instead of button's bg/fg

actionless:
(in some GTK themes firefox not detects the ones for button correctly)


piroor:
Thanks. And I've applied one more change to make "inverted" color surely visible.

PR #1720, opened 2018-01-18
German localization

sicherist:
I created a German translation. I had to base the translation on the English locale, as I don't speak Japanese unfortunately, so there may be some inaccuracies. Obviously I am not using all of the many features of TST, so I hope that all translations are correct in context.

What I didn't find in the messages.json is this description:

screenshot_20180117_210319

I could also translate this if you like and tell me where to find it.

### DIFF
--- a/webextensions/_locales/de/messsages.json
+++ b/webextensions/_locales/de/messsages.json
@@ -1,0 +1,282 @@
+{
+  "extensionName": { "message": "Tree Style Tab" },
+  "extensionDescription": { "message": "Stellt Tabs in einer Baumstruktur dar." },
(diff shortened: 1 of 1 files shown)


piroor:
Thanks a lot!

Descriptions at the addon page are not managed under this repository yet. For easy translation I need to import them to this repository at first.

piroor:
I've added descriptions on AMO website to the directory:
https://github.com/piroor/treestyletab/tree/master/docs/addons.mozilla.org

Please copy non-suffixed files (English) to `.de`-suffixed and translate them.

PR #1721, opened 2018-01-19
German translation of AMO description

sicherist:
Here is the german translation of the AMO description files based on the english version. I must say that TST is way more powerful than this humble description makes it seem :)


piroor:
Thanks a lot!

Issue #1714, opened 2018-01-11
White flash when loading sidebar


overdodactyl:
When the sidebar loads, the background color is a bright white (either when opening a new window, or toggling Tree Style Tabs open).

When opening the sidebar, the progression is as follows:

Complete white background --> White Background with a white "new tab button"  -->  dark background with loading dots -> custom styling + tabs

Here's a gif (sorry for the poor quality, I'm not sure why the streaking is showing up):

![white_flash](https://user-images.githubusercontent.com/33071129/34839104-6992ccb8-f6f8-11e7-94a4-25bc109b724a.gif)

Is there any way to change the background color used here?

I use 

```css
/* Remove white flash on load */
#sidebar, #sidebar-box {
    background: var(--primary-dark-color) !important;
}
```

to remove the white flash from FF's sidebar options (history, bookmarks, etc.), but couldn't figure anything out for TST.

Thanks!

## Environment

 * Platform (OS): macOS
 * Version of Firefox: 59
 * Version (or revision) of Tree Style Tab: 2.4.3

piroor:
Because addon's sidebar panel is implemented by deeply nested inline frames, you need to apply background color for more elements:

```
<browser id="sidebar"> <= this is already applied
  #document at chrome://browser/content/webext-panels.xul?...
    <page id="webextpanels-window"> <= not applied yet
      <stack>
        <browser id="webext-panels-browser"> <= not applied yet
          #document at moz-extension://9071bd23-5e4b-4582-a58b-0ac99c1256b6/sidebar/sidebar.html?...
          ...
        </browser>
      </stack>
    </page>
</browser>
```

BTW, after recent changes now TST applies its custom background color to its sidebar panel ASAP. I think this change should reduce this flashing.

overdodactyl:
Hi @piroor, thank you so much for the help and adjustments to TST, it's really appreciated!

I've added the additional elements from the nested inline frames and that seems to have effectively removed the initial white background seen before TST's white background + new tab button is seen.  

I think that alongside TST applying the background at an earlier stage should completely resolve the flashing.

Thanks again!

overdodactyl:
Here are the relevant different frames from TST upon loading (the black is from the default dark theme available, then the dark grey is my custom styling):

![frames](https://user-images.githubusercontent.com/33071129/34893219-2aaa0ca0-f7d4-11e7-9895-9f070ee8482c.jpg)

[...]

PR #1729, opened 2018-01-24
Add defer attribute to script elements

saneyuki:
## Motivation

`script[type="module"]` will be executed on when parsing html document is complete. This behavior is a similar to the one of `script[defer]`.

This means that `script[defer]` decreases a complexity to migrate to a module script from a classic script.

## See also

- [4.12.1 The script element - WHATWG HTML Standard](https://html.spec.whatwg.org/multipage/scripting.html#the-script-element)

### DIFF
--- a/webextensions/background/background.html
+++ b/webextensions/background/background.html
@@ -5,31 +5,31 @@
 <html>
   <head>
     <meta charset="UTF-8">
-    <script type="application/javascript" src="/common/tree/constants.js"></script>
-    <script type="application/javascript" src="/common/fake-context-menu/constants.js"></script>
-    <script type="application/javascript" src="/common/fake-context-menu/background.js"></script>
-    <script type="application/javascript" src="/common/Configs.js"></script>
-    <script type="application/javascript" src="/common/common.js"></script>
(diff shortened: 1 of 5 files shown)


saneyuki:
@piroor how about do you think?

piroor:
Sadly this change reintroduces the problem reported at #1714. Do you have any good idea to  solve both?

saneyuki:
@piroor 

I'm not sure about the root problem with #1714. I suspect that scripts will not block to parse a document and sets the style just before `DOMContentLoaded`. Even if we switch to a module script, #1714 would be reproducible.

I feel it may fix to apply dark background color to body or document style because user attention will focus to a white color.

saneyuki:
@piroor 

I rebased this pull request on the tip of `master` branch.
And I had tried whether this change still causes the regression about #1714, but I didn't reproduce it by the rebased this branch. Perhaps my reproduce steps was wrong.

How do you feeling about this?

piroor:
Hmm, the regression I worried seems not happen for me also. And I misunderstood that `defer` as `async`. I've decided to merge this and try further more doogfooding.

piroor:
Anyway, thanks!

PR #1738, opened 2018-01-26
Update zh_TW locale and remove needless blankline in ja locale

lycsjm:
That blankline don't appear in en and old zh_TW locale, so I ust deleted it for unify three locale.


piroor:
Thanks a lot!

PR #1759, opened 2018-02-05
Update german locale

sicherist:
Updated the german translation.

I have a question about something that would make it way easier for me to maintain the german translation: Would it be possible that you would add future changes you do to "webextensions/_locales/**en**/messages.json" also to  "webextensions/_locales/**de**/messages.json"? For example when you are done editing the English messages.json just copy the english stuff to the german messages.json? This way I could immediately see when something has changed and quickly translate it to german.

Best regards!


piroor:
Thank you for updating!

Adding new entries in English to each localized resources will introduce another problem: I have to update all untranslated messages in each locale when I change existing messages. On the other hand, WebExtensions' i18n feature automatically fallbacks to the entry defined in the default locale if the entry is missing in the current locale (see also: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Internationalization ) thus we don't need to add same unlocalized message to each locale. This is the largest reason why I maintain only en and ja locales.

I think that something helper tools for localization (or internationalization) should solve this problem. For example there is a plugin for the Sublime Text: https://github.com/Harurow/sublime_chromeextensioni18nhelper

Or, a diff between latest tag and the `master` branch will also help you to see difference of `_locales/en/messages.json` like: https://github.com/piroor/treestyletab/compare/2.4.8...master

piroor:
There is another utility: https://www.npmjs.com/package/web-ext-translator

sicherist:
Thanks, I'll have a look at the tools you recommended.

PR #1863, opened 2018-04-12
Fix incorrect tab tracking

### LINKED ISSUE #1709: TST misplaces new tabs that open in a different container
## Steps to reproduce

 1. Set up a site (e.g. github.com) to always open in a specific container
 2. In a different container (or no container), open at least two tabs
 3. In the first tab, middle-click a link to github.com to open it as a new tab

## Expected result

Github link opens as a child of first tab

## Actual result

(See screenshot)

Github link opens as a child of the second tab. The tabbing order is maintained, however. i.e. in the screenshot you can see that "next tab" cycles through tabs as if they were in the correct placement, with the github link activating after the first tab, and before the second (even though it's rendered as a child of the second).

If the second tab already has a child (as in screenshot), the animation hints that the new tab is opened in the right place, then quickly gets closed and reopened in the wrong place. I'm assuming that's to do with how firefox opens a tab in a different container.

## Environment

 * Platform (OS): Fedora
 * Version of Firefox: 57.0.1
 * Version (or revision) of Tree Style Tab: 2.4.3

screencast from 06-01-18 22_10_41

This may be related to other issues about opening tabs rapidly, but it's hard to be sure, and none of them mentioned containers.

Lej77:
Since created tabs are positioned after an await the tab order can be broken if the order is manipulated during the await time. This merge request moves the positioning code to before all awaits. This seems to fix #1709.

The [test script in #1709](https://github.com/piroor/treestyletab/issues/1709#issuecomment-358880326) doesn't break the tab order after this fix but does before. This means that the `Always open in [Container]` option in the popup for [Firefox Multi-Account Containers by Mozilla](https://addons.mozilla.org/firefox/addon/multi-account-containers/) doesn't break the tab order after this fix.

### DIFF
--- a/webextensions/common/tree/handlers.js
+++ b/webextensions/common/tree/handlers.js
@@ -225,17 +225,14 @@ async function onNewTabTracked(aTab) {
 
   var newTab = buildTab(aTab, { inRemote: !!gTargetWindow });
   newTab.classList.add(kTAB_STATE_OPENING);
-  // append to DOM tree to detect duplication
-  container.appendChild(newTab);
+
+  var nextTab = getAllTabs(container)[aTab.index];
+  container.insertBefore(newTab, nextTab);
 
   gCreatingTabs[aTab.id] = newTab.uniqueId;
   var uniqueId = await newTab.uniqueId;
   if (gCreatingTabs[aTab.id] === newTab.uniqueId)
     delete gCreatingTabs[aTab.id];
-
-  // move to correct position after tabs.onRemoved is processed
-  var nextTab = getAllTabs(container)[aTab.index];
-  container.insertBefore(newTab, nextTab);
 
   updateTab(newTab, aTab, {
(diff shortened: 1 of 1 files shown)


blackhole89:
As far as I can tell, it fixes the problem of FF tab order (as observed by ctrl-tab) and display order in TST being inconsistent, but introduces a new one: if the multi-account containers confirmation page that is replacing itself with a new tab in a different container is the bottommost (last) tab of a subtree, then the replacement tab is lifted to the top level. (This can result in an unnatural tree state, where a tab without parents is surrounded on both sides by tabs that are under some other tab further up.)

piroor:
@Lej77 Thanks. Finally I've realized that there is less reason to prevent putting a new tab to its correct position at first, so I've merged your PR.

@blackhole89 I think you should open another new issue to track regressions.

PR #1864, opened 2018-04-12
Separate different button's mouse events.

Lej77:
Improves mouse event management for the Sidebar. Currently data is gathered at the mouse down event and that data is later used on the mouse up event. But if another mouse button is pressed without releasing the first mouse button then the info gathered can be about the wrong mouse button.

For example:
 1. Press and hold the left mouse button over a tab.
 2. Press and hold the middle mouse button over the same tab.
 3. Release the left mouse button over the same tab.

Expected Result:
The tab should remain open and only close on release of middle click.

Actual Result:
The tab is closed.

### DIFF
--- a/webextensions/sidebar/handlers.js
+++ b/webextensions/sidebar/handlers.js
@@ -158,10 +158,10 @@ function onContextMenu(aEvent) {
   });
 }
 
-var gLastMousedown = null;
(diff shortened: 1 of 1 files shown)


piroor:
Thanks! This PR was conflicted to my recent changes, so I've merged by hand.

PR #1865, opened 2018-04-12
[API Improvement] Allow preventing middle click to close tab 

### LINKED ISSUE #1861: [API Improvement] Allow preventing middle click to close tab
## Short description
Before Firefox 57 I was using Tree Style Tab together with [Auto Unload Tab by def00111](https://addons.mozilla.org/firefox/addon/auto-unload-tab/). I configured Auto unload tab to unload tabs when they were middle clicked, so tabs were only closed on middle click if they were already unloaded. I tried to replicate this behavior with my extension [Tab Unloader for Tree Style Tab by Lej77](https://addons.mozilla.org/firefox/addon/tab-unload-for-tree-style-tab/). Unfortunately I can't prevent Tree Style Tab from closing a tab when it is middle clicked. It would be nice if this could be implemented.

## Steps to reproduce

 2. Install TST.
 3. Install [Tab Unloader for Tree Style Tab by Lej77](https://addons.mozilla.org/firefox/addon/tab-unload-for-tree-style-tab/).
 4. Middle click a loaded tab.

## Expected result
The tab should be unloaded but remain open.

## Actual result
The tab is closed.

## Environment

 * Platform (OS): Windows 10
 * Version of Firefox: Firefox 59.0.2
 * Version (or revision) of Tree Style Tab: 2.4.18

Lej77:
Allow preventing middle click to close tab by sending true when listening to `mouse-up` TST API event.
This will fix #1861.

### DIFF
--- a/webextensions/sidebar/handlers.js
+++ b/webextensions/sidebar/handlers.js
@@ -298,8 +298,9 @@ async function onMouseUp(aEvent) {
   let tab = getTabFromEvent(aEvent);
 
   let serializedTab = tab && serializeTabForTSTAPI(tab);
+  let results;
   if (serializedTab && gLastMousedown) {
-    sendTSTAPIMessage(Object.assign({}, gLastMousedown.detail, {
+    results = sendTSTAPIMessage(Object.assign({}, gLastMousedown.detail, {
       type:    kTSTAPI_NOTIFY_TAB_MOUSEUP,
       tab:     serializedTab,
       window:  gTargetWindow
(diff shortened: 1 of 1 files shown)


piroor:
I think only the "detection of canceled" state should be exposed instead of the raw responses of `tab-mouseup` message. The commit 58df17336e16f175dedee73685b0e9e9feeb4e36 should cover this PR, thus I close this. Anyway, thank you for the advice!

Lej77:
@piroor Your code always uses `await` on the results which means that functionality that shouldn't be affected by that wait is. Just imagine if there is an infinite wait at that point. What functionality could that break? My implementation only waited on the results just before closing the tabs and the `onMouseUp` function could continue to execute its code even when we are waiting for the results.

piroor:
Thanks! I missed the viewpoint... By the commit 413a614 TST doesn't wait responses from other extensions until the information is really required.

Lej77:
You new commit seems fine. 

At first I thought that a delayed API promise could cause a API message for `kTSTAPI_NOTIFY_TABBAR_MOUSEUP` to be sent long after a click happened but after looking some more I realized that `promisedCanceled` will only be a promise set by the API if the `tab` variable is true and under those circumstances the code on lines 381-395 won't be triggered. This might be clearer if the `!(await promisedCanceled)` on line 354 is changed to `!tab || !(await promisedCanceled).

piroor:
By recent commits, now the function `onMouseUp` is rewritten to early-return style, so I think it became more readable now.

Lej77:
Yes, that is a lot better!

Issue #1803, opened 2018-02-22
Focus should change on mouseUp so that we can move tabs without loading them


ariasuni:
## Steps to reproduce

 1. Start Firefox with clean profile.
 2. Install TST.
 3. Open a few tabs 
 4. Check “Show your windows and tabs from last time” on Firefox settings page
 5. Restart Firefox
 6. Try to move unloaded tabs 

## Expected result

I can’t drag and drop tabs without the focus changing and Firefox loading them.

## Actual result

Tab is loaded as soon as I click on it, so I can’t move it without loading it.

## Environment

  * Platform (OS): GNU/Linux
  * Version of Firefox: 59.0b11
  * Version (or revision) of Tree Style Tab: 2.4.16

piroor:
I don't think this is a "bug" of TST, because this is quite same behavior to Firefox's horizontal tab bar. I think TST should respect Firefox's behavior when it is possible - this is a design policy. Sorry.

alreadydone:
So is this bug on Bugzilla now? If not we could provide this through an option of TST as @Lej77 already has a patch for it.

piroor:
After the commit d460cea, now extensions like described at https://github.com/piroor/treestyletab/pull/1866 can work as expected. Thus it should become a workaround for this issue.

@alreadydone If you think dragged tabs should keep suspended on Firefox's horizontal tab bar, you should file a bug for it. If Firefox's behavior is changed in a future versions of Firefox, I will change TST's behavior to follow the new one of Firefox itself.

Lej77:
I released an extension based on my example code on `addons.mozilla.org`. So that extension should work as soon as the next version of Tree Style Tab releases.

Here is a link: [Move unloaded tabs for Tree Style Tab by Lej77](https://addons.mozilla.org/firefox/addon/move-unloaded-tabs-for-tst/)

QuixoticGitHub:
This needs to be resurrected. Loading onmousedown makes absolutely no sense. If it's a "design philosophy" then you need to reevaluate your axioms.

ariasuni:
If you don’t like it, you can just use [Lej77’s addon](https://github.com/piroor/treestyletab/issues/1803#issuecomment-382371644).

Even if I don’t agree it’s the best behavior, sticking to Firefox’s default behavior has its advantages.

Furthermore, from the [README.md](https://github.com/piroor/treestyletab#faq--frequently-rejected-requestsproposals):

[...]

PR #1866, opened 2018-04-12
Option to not focus on moved tabs

Lej77:
I wanted to address #1803. I think the best solution is to use an extension to prevent the default action for the left mouse button if the tab is dragged. I made an extension that did this and I added the code bellow but this didn't work since Tree Style Tab always selects tabs when they are moved. So I made this merge request to add an option in TST to disable that behavior.

## Example extension

`background.js`:
```JavaScript
const kTST_ID = 'treestyletab@piro.sakura.ne.jp';

async function registerToTST() {
  try {
    await browser.runtime.sendMessage(kTST_ID, {
      type: 'register-self',
      name: browser.runtime.id,
      listeningTypes: ['tab-mousedown', 'tab-mouseup'],
    });
  } catch (error) { return false; }
  return true;
}

var lastResolve = null;
function resolveAs(value) {
  if (lastResolve && typeof lastResolve === 'function') {
    lastResolve(value);
  }
  lastResolve = null;
}

var lastMessage;
browser.runtime.onMessageExternal.addListener((aMessage, aSender) => {
  if (aSender.id !== kTST_ID) {
    return;
  }
  switch (aMessage.type) {
    case 'ready':
      registerToTST(); // passive registration for secondary (or after) startup
      break;
    case 'tab-mousedown':
      if (aMessage.button !== 0) {
        return false;
      }
      return new Promise((resolve, reject) => {
        resolveAs(true);
        lastResolve = resolve;
        lastMessage = aMessage;
      });
      break;
    case 'tab-mouseup':
      if (aMessage.button !== 0) {
        return false;
      }
      let releasedWithoutMove = lastMessage && aMessage.tab.id === lastMessage.tab.id;
      resolveAs(!releasedWithoutMove);
      break;
  }
});
registerToTST(); // aggressive registration on initial installation
```

`manifest.json`:
```JSON
{
  "manifest_version": 2,
  "name": "Move unloaded tabs for Tree Style Tab",
  "version": "1.0",
  "applications": {
    "gecko": {
      "strict_min_version": "57.0"
    }
  },
  "background": {
    "scripts": [
      "background.js"
    ]
  }
}
```

### DIFF
--- a/webextensions/common/common.js
+++ b/webextensions/common/common.js
@@ -225,6 +225,8 @@ configs = new Configs({
   minimumIntervalToProcessDragoverEvent: 50,
   moveDroppedTabToNewWindowForUnhandledDragEvent: true, // see also: https://github.com/piroor/treestyletab/issues/1646
   autoDiscardTabForUnexpectedFocus: false,
+  focusOnMovedTabs: true,
+  focusOnMovedTabsBetweenWindows: true,
   knownExternalAddons: [
     'multipletab@piro.sakura.ne.jp'
   ],
--- a/webextensions/common/tree/tree-operations.js
+++ b/webextensions/common/tree/tree-operations.js
@@ -1475,8 +1475,14 @@ async function performTabsDragDrop(aParams = {}) {
   else
     log('=> already placed at expected position');
 
-  browser.tabs.update(draggedTabs[0].apiTab.id, { active: true })
-    .catch(handleMissingTabError);
+  if (
+    windowId !== destinationWindowId ?
+    configs.focusOnMovedTabsBetweenWindows :
+    configs.focusOnMovedTabs
+  ) {
+    browser.tabs.update(draggedTabs[0].apiTab.id, { active: true })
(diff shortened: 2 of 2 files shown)


piroor:
You don't need this change after d460cea. Now the dragged tab is not activated after dropped. Originally the code was introduced to make the dropped tab visible after it is attached to a collapsed tree, and the behavior has been re-implemented in a different way.

By the way, basically I think we should not add new "option" except when it is really needed. Too many optional choices will make people confused, so I think we should always find ways to solve a problem without high cost.

Lej77:
I released an extension based on my example code on `addons.mozilla.org`. So that extension should work as soon as the next version of Tree Style Tab releases.

Here is a link: [Move unloaded tabs for Tree Style Tab by Lej77](https://addons.mozilla.org/firefox/addon/move-unloaded-tabs-for-tst/)

piroor:
Good news! I'm using your new extension now.

PR #1867, opened 2018-04-12
Readme update - Add extensions to the list of related addons

Lej77:
Add more extensions that use the Tree Style Tab API to the readme. Note that I don't know if the `Tree Style Tab Focus Preceding Tab on Close` extension is released on AMO, I couldn't find it when I searched for it. I included it anyway since the code might be helpful to other users and it is would be hard to find if not included.


piroor:
Thanks！

PR #1875, opened 2018-04-21
Context menu item for Page Action to open settings page

Lej77:
I saw that you could have context menu items for an extensions browser action. I think it would be nice to use this to provide rarely used context menu items such as one to open the extensions option page. 

This merge request adds the ability to quickly open the Tree Style Tab settings page by right clicking the extensions browser action (the button in the toolbar) and selecting the right context menu item.


piroor:
Sorry, I didn't add this feature intentionally so this PR won't be merged. My policies about this feature are:

 * Addon should work smartly, without user's configuration.
 * When the user have to change configurations again and again, something wrong. Easy access to the configuration page actually helps such users, but it may leave real problems there.

PR #1898, opened 2018-05-20
Readme update - Add extension to the list of related addons

### LINKED ISSUE #1756: [Feature Request] Bookmark and restore tree information
## Store Tree information in bookmarks

To store meta information in bookmarks extension's currently have three options that I can see:
1. Store it in a bookmark's title.
2. Store it in a bookmark's URL.
3. Store it with the bookmarks structure (folders and separators).

For the first option I was thinking that tree information could be stored in a bookmark's title by appending a tab-id and a parent-tab-id. Something like: 
`Google [Tab-id: 185875] [Parent-id: 184321]`

For the second option the tree information could be encoded in a custom URL that also contained the original URL. Something like: 
`about:treestyletab-bookmark?tab-id=185875&parent-id=184321&url=www.google.com`

For the third option the tree information could be stored by putting all child tabs for a tab in a sub-folder. Something like:
[Google](www.google.com) (parent)
`Google` (folder)
|- [Tree Style Tab – Add-ons for Firefox](https://addons.mozilla.org/en-US/firefox/addon/tree-style-tab/ "Tree Style Tab – Add-ons for Firefox") (child of: Google)
|- `Tree Style Tab – Add-ons for Firefox` (folder)
| |- [piroor/treestyletab: Tree Style Tab, Show tabs like a tree.](https://github.com/piroor/treestyletab "piroor/treestyletab: Tree Style Tab, Show tabs like a tree.") (child of: Tree Style Tab – Add-ons for Firefox)
| |- `piroor/treestyletab: Tree Style Tab, Show tabs like a tree.` (folder)
| | |- [Issues · piroor/treestyletab](https://github.com/piroor/treestyletab/issues "Issues · piroor/treestyletab") (child of: piroor/treestyletab: Tree Style Tab, Show tabs like a tree.)
| | |- [Commits · piroor/treestyletab](https://github.com/piroor/treestyletab/commits/master "Commits · piroor/treestyletab") (child of: piroor/treestyletab: Tree Style Tab, Show tabs like a tree.)

## Restore tree from bookmarks

For the second method this is quite obvious, just parse the tree information from the URL and then redirect the page to the correct URL. While this method is straightforward to implement it has the disadvantage that the bookmarks will stop working if TST is uninstalled since they will then stop being redirected. This could be somewhat mitigated by having an option to transform all such TST bookmarks to normal bookmarks in TST settings page or something similar.

The first methods require that the extension knows from which bookmarks the tabs were opened from. This could probably in most cases be accomplished by searching all bookmarks but that isn't really a feasible solution since that would be very slow and there might be several possible results from such a search. This would also not help with restoring a tree saved with the third method. 

I see two solutions to this, either a special bookmark is generated at the top of the bookmark folder that should have a unique URL that the extension can search for. Since a bookmark search can take a while this still has the disadvantage of being very slow though. The other solution is to get the relevant bookmarks' locations with the help of a context menu option. 

In the blog post [Extensions in Firefox 59 | Mozilla Add-ons Blog](https://blog.mozilla.org/addons/2018/01/26/extensions-firefox-59/ "Extensions in Firefox 59 | Mozilla Add-ons Blog") under the header `Context Menus for Bookmarks` it is stated that context menu items for bookmark will work for extensions in Firefox 59. It doesn't work for the bookmark sidebar or the bookmark library window but it does work for the bookmark button you can add from the customize page. It also work with the bookmark toolbar and the Hamburger menu -> Library ->  Bookmarks menu. 

This means that in Firefox 59 and later we could have a custom context menu alternative for bookmarks. Something like: "Restore TST Tree". This could be used to restore a tree from any of three storage options suggested above.

I think the first method is best since the second and third methods makes the bookmarks hard to use without having TST installed.

Lej77:
I made an extension to bookmark and restore trees to fix #1756. Would be nice to include it in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,7 @@ Some addons provide extended behavior to TST's sidebar panel:
  * [Tree Style Tab in Separate Window](https://addons.mozilla.org/firefox/addon/tst-in-separate-window/) allows you to open the Tree Style Tab sidebar page in a new window.
  * [Auto Tab Discard](https://addons.mozilla.org/firefox/addon/auto-tab-discard/) supports the fake context menu in the Tree Style Tab sidebar.
  * [UnloadTabs](https://addons.mozilla.org/firefox/addon/unload-tabs/) supports the fake context menu in the Tree Style Tab sidebar.
+ * [Bookmark Tree for Tree Style Tab](https://addons.mozilla.org/firefox/addon/bookmark-tree-for-tst/) allows you to bookmark and restore trees.
 
 ## Similar projects
 


piroor:
Thanks!

PR #1925, opened 2018-06-08
Change the counter color which is black

Nicolas01:
The counter at the right of each tab label is in black, so it is difficult to read it.
I changed the counter color to be the same as the label.

### DIFF
--- a/webextensions/sidebar/styles/base.css
+++ b/webextensions/sidebar/styles/base.css
@@ -361,6 +361,14 @@ ul {
   content: ")";
 }
 
+/* change the counter color which is black */
+.tab.discarded .counter {
+  color: var(--tone-3)!important
+}
+.tab .counter {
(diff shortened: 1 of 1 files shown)


piroor:
Who defines `--tone-1` and `--tone-3`? This is the first time I see them and they look not defined by Firefox itself: https://dxr.mozilla.org/mozilla-central/search?q=tone-1+file%3A.css&redirect=false

And, could you attach a screenshot to here to indicate what is the problem? Currently I cannot imagine how current style is broken on you environment.

Nicolas01:
Sorry, `--tone-1` and `--tone-3` come from [ShadowFox ](https://github.com/overdodactyl/ShadowFox) which is a dark theme.
I mixed up everything, my fix should go to ShadowFox instead of treestyletab.

PR #1927, opened 2018-06-10
Convert background/context-menu.js into ES Module 

saneyuki:
This pull requests contains these changes:

1. Allow to add es module style file to background module.
2. Enable `no-undef` rule for them.
3. Try to split `gMetricsData` into own module.
4. Convert background/context-menu.js into ES Module.

### DIFF
--- a/webextensions/background/module/index.js
+++ b/webextensions/background/module/index.js
@@ -1,4 +1,10 @@
+import {
+  refreshContextMenuItems,
+  contextMenuClickListener,
+} from './context-menu.js';
 import { gMetricsData } from './metrics.js';
 
 // Set to the global to make compatibility with other classic sources.
 window.gMetricsData = gMetricsData;
+window.refreshContextMenuItems = refreshContextMenuItems;
+window.contextMenuClickListener = contextMenuClickListener;
--- a/webextensions/background/module/index.js
+++ b/webextensions/background/module/index.js
@@ -1,0 +1,4 @@
+import { gMetricsData } from './metrics.js';
+
+// Set to the global to make compatibility with other classic sources.
+window.gMetricsData = gMetricsData;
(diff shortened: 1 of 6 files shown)


saneyuki:
@piroor 

How do you think about this?

webextensions/background/module/index.js
```diff
@@ -0,0 +1,10 @@
+import {
+  refreshContextMenuItems,
+  contextMenuClickListener,
```
piroor: There is no need to define these functions as global functions, and I think they should be enclosed under a namespace like: `contextMenu.refreshItems` and `contextMenu.onClick`.

webextensions/background/module/metrics.js
```diff
@@ -0,0 +1,5 @@
+// Defined in a classic script source, and we can read these as global variables. 
+/* global
+  MetricsData: false,
+ */
+export const gMetricsData = new MetricsData();
```
piroor: Why don't you separate this change as another PR? This PR looks doing too much things. Focusing to one topic for each PR, please.

piroor:
Thanks. I've added some comments.

webextensions/background/module/index.js
saneyuki: > There is no need to define these functions as global functions, and I think they should be enclosed under a namespace like: contextMenu.refreshItems and contextMenu.onClick.

To not change usages in `background/background.js`, I didn't assign to namespace object.
Should we change them?

webextensions/background/module/metrics.js
saneyuki: I create https://github.com/piroor/treestyletab/pull/1927/commits/5c737d8eeedd41b49624d12f41e873d5f03c1296 as minimal PoC change to convert es module.

> Focusing to one topic for each PR, please.

Okay. I'll create a new pull request.

saneyuki:
@piroor 

I will close this pull request once, and create pull requests which resolves each of topics.

saneyuki:
I created https://github.com/piroor/treestyletab/pull/1929 as retaking this.

PR #1928, opened 2018-06-10
Allow to add ES Module file to **/module/ 

saneyuki:
* This retakes https://github.com/piroor/treestyletab/pull/1927
* This intend to create the directory to contain an ES Module.


piroor:
Thanks. I hope that most JS codes become modules and these "module" directory disappear in the future. (Is that possible?)

saneyuki:
> Thanks. I hope that most JS codes become modules and these "module" directory disappear in the future. (Is that possible?)

I think so. We can remove _module_ directory by flatten each of directories after we convert all files into ES Module.

PR #1929, opened 2018-06-10
Convert background/context-menu.js into ES Module 

saneyuki:
* This retakes #1927
* This depends on #1928

### DIFF
--- a/webextensions/background/module/index.js
+++ b/webextensions/background/module/index.js
@@ -1,0 +1,8 @@
+import {
+  refreshContextMenuItems,
+  contextMenuClickListener,
+} from './context-menu.js';
+
+// Set to the global to make compatibility with other classic sources.
+window.refreshContextMenuItems = refreshContextMenuItems;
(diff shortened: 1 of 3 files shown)


webextensions/background/module/index.js
```diff
@@ -0,0 +1,8 @@
+import {
+  refreshContextMenuItems,
+  contextMenuClickListener,
+} from './context-menu.js';
+
+// Set to the global to make compatibility with other classic sources.
+window.refreshContextMenuItems = refreshContextMenuItems;
+window.contextMenuClickListener = contextMenuClickListener;
```
saneyuki: @piroor 

I think [this discussion](https://github.com/piroor/treestyletab/pull/1927#pullrequestreview-127403079) is continuing.
Should I change callers of these?

piroor: I think that such too-long-name functions are not module-friendly, so they should be renamed simply and other codes should touch them as methods of a named module (as an enclosed namespace.)

Anyway, current code looks good to do things incrementally.

PR #1933, opened 2018-06-10
Enable ESLint rules detecting unused variables for ES Module files

saneyuki:
* This depends on https://github.com/piroor/treestyletab/pull/1932
* By this change, we can detect unused variables.
   * Under ES Module, we should/can clarify where a variable defines. Introducing this is easier.


piroor:
Thanks! This will help us to write code more safely and cleanly.

Issue #1935, opened 2018-06-11
[meta] Migrate to ES Module 


saneyuki:
This is proposal to migrate to ES Module.

## Motivation

* ES Module is a new standard way to organize a JS code.
* We can use it from Firefox 60.

### 0. Change the way loading scripts to `defer`

This is done by https://github.com/piroor/treestyletab/pull/1729

### 1. Allow to place ES Module and load them

This is done by https://github.com/piroor/treestyletab/pull/1928

### 2. Enable more ESLint rules to detect

* This is still progress and we can proceed next steps.
   * https://github.com/piroor/treestyletab/pull/1933
   * Add [eslint-plugin-import](https://github.com/benmosher/eslint-plugin-import).
   * and more....?
* In ES Module, we can make where all variables is defined statically in syntax. So we can detect unused variables or invalid importing by ESLint or other static analysis. It' easier to make our development.

### 3. Convert classic script source into ES Module.

* This is still progress
   * https://github.com/piroor/treestyletab/pull/1929
   * https://github.com/piroor/treestyletab/pull/1930
   * https://github.com/piroor/treestyletab/pull/1931

### 4. Flatten directories.

After complete step 3, then we would have only ES Modules!
We can erase `**/modules/` introduced by _step 1_ by flatten them.

saneyuki:
@piroor 

This is my big picture. How about do you think?

piroor:
Thank you for the plan. Experimentally I converted more independent modules to ES modules, and I've understood that more aggressive linting helps our development.

I think there are more detailed steps in the section "3. Convert classic script source into ES Module." (we are here):

 1. Convert base, get-tabs, tab-information, and tree-operations (possibly, more scripts) to ES modules.
 2. Restructure modules created at the step 1 to two ES modules: `Tabs` and `Tree`. `Tabs` should provide generic operations around tab elements in TST, `Tree` should provide methods to operate tree of tabs.
 3. Convert rest modules depending on `Tabs` and `Tree` to ES modules.
 4. Convert external libraries to ES modules. (This can be done at first.)

How do you think this detailed plan, especially the step 2? I think it is necessary to solve current too-complex and too-fragmented scripts, but it will take more time.

piroor:
Another plan is: doing the step 2 after we completely migrate to ES modules. Because the step 2 will break everything easily, so aggressive linting introduced by completely-modulized codes will help that.

piroor:
[...]

PR #1937, opened 2018-06-11
Enable import/no-unresolved to detect to import from no-exist files

saneyuki:
By this rule, we can detect typical misloading 😄 

## Related Issues

https://github.com/piroor/treestyletab/issues/1935


piroor:
Thanks!

PR #1939, opened 2018-06-11
Update ESLint to the latest to allow ES2018 syntax 

saneyuki:
By this change, we can these new syntax features.

* [object rest/spread properties](https://github.com/tc39/proposal-object-rest-spread) (at least Firefox 59~)
* [async iteration](https://github.com/tc39/proposal-async-iteration) (at least Firefox 59~)
   * `for await (const item of async_iterable) {}`
   * `async function* asyncGenerator() { yield }`

I'm not sure about that we already have a code which we wait to use them. But these may be useful for the future.

### DIFF
--- a/webextensions/.eslintrc.js
+++ b/webextensions/.eslintrc.js
@@ -10,7 +10,7 @@ module.exports = {
   'root': true,
 
   'parserOptions': {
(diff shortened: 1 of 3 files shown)


piroor:
Thanks! It will help future development.

PR #1940, opened 2018-06-11
Enable import/no-absolute-path to disallow using absolute path to import an ES Module

saneyuki:
This enables [import/no-absolute-path]( https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-absolute-path.md).

By this change, ESLint enforce us to use a relative path.

Now, we use a relative path to empower ESLint's static analysis without setting eslint-plugin-import's _resolver_. So I think we can enable this rule to ensure that ESLint can always empower us to development easily.

### DIFF
--- a/webextensions/tools/eslint/for-module.js
+++ b/webextensions/tools/eslint/for-module.js
@@ -41,6 +41,10 @@ module.exports = {
     'import/default': 'error',
     // https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/namespace.md
     'import/namespace': 'error',
+    // https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-absolute-path.md
+    'import/no-absolute-path': ['warn', {
+      'esmodule': true,
+    }],
     // https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-duplicates.md
     'import/no-duplicates': 'error',
     // https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/export.md


piroor:
Hmm, I'm negative to enforcing this rule because importing with relative path looks just a hack for restriction of the linter. In other words, code in addon should be designed for running on the browser initially, and the linting rule should be designed to fit to the final running environment. If resolver for eslint-plugin-import can be written with short code, I think it is better than modifying addon's source.

saneyuki:
@piroor 

> Hmm, I'm negative to enforcing this rule because importing with relative path looks just a hack for restriction of the linter. In other words, code in addon should be designed for running on the browser initially, and the linting rule should be designed to fit to the final running environment. If resolver for eslint-plugin-import can be written with short code, I think it is better than modifying addon's source.

I think we should enable this once by these reasons:

1. At now (https://github.com/piroor/treestyletab/commit/01721259719842a5e014e442257ecdfdaf080339), we have already used an absolute path.
2. Until changing the configuration to support an absolute path, we need to write a relative path. We should support its situation by linting.
3. A web browser also supports a relative path to import an ES Module.

For the future, we may switch to use an absolute path to specify a path to import. Then we should switch all of path at once. Now, I don't think it's the time to do so [by your comment](https://github.com/piroor/treestyletab/issues/1935#issuecomment-396293732). I seem that it would cause many conflicts if we switch to using an absolute path. At least, we need to use a relative path until converting all scripts to ES Module.

I filed https://github.com/piroor/treestyletab/issues/1942 for the future.

piroor:
This became obsolete by fixed #1942.

PR #1946, opened 2018-06-13
Copy external libraries into external/ 

saneyuki:
## Motivation

It's hard to recognize that it's an external libraries a file which is in managed directories by git.

## Detailed Design

* Copy external libraries into `external/`
   * We may have any other good naming for this library. But I think this is most straight and reasonable at now.

### DIFF
--- a/webextensions/background/background.js
+++ b/webextensions/background/background.js
@@ -5,6 +5,9 @@
 */
 'use strict';
 
+import TabFavIconHelper from '../external/TabFavIconHelper.js';
+import RichConfirm from '../external/RichConfirm.js';
+
 import {
   log,
   dumpTab,
@@ -25,8 +28,6 @@ import * as TSTAPI from '../common/tst-api.js';
 import * as SidebarStatus from '../common/sidebar-status.js';
 import * as Commands from '../common/commands.js';
 import * as Migration from '../common/migration.js';
-import TabFavIconHelper from '../common/TabFavIconHelper.js';
-import RichConfirm from '../common/RichConfirm.js';
 import EventListenerManager from '../common/EventListenerManager.js';
 
 import * as TreeStructure from './tree-structure.js';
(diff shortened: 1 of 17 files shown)


webextensions/background/background.js
```diff
@@ -250,7 +251,7 @@ export async function tryInitGroupTab(aTab) {
     return;
   browser.tabs.executeScript(aTab.apiTab.id, Object.assign({}, scriptOptions, {
     //file:  '/common/l10n.js'
-    file:  '/resources/l10n.js' // ES module does not supported as a content script...
+    file:  '/external/l10n-classic.js' // ES module does not supported as a content script...
```
saneyuki: I use _classic_ as the term of traditional script type by [_classic script_](https://html.spec.whatwg.org/multipage/webappapis.html#classic-script) in whatwg html spec

piroor:
This change looks good!

There is one concern: how should we name the directory? For example, Ruby on Rails uses "vendor" for the purpose. I have no knowledge about such a naming on JS products, so I don't know "external" is major or not. The current name "external" looks enough reasonable so I agree to use the name if there is no other known major naming for this purpose. Any advice?

saneyuki:
@piroor 

I chose "external" by these reason:

* I seem _vendor_ means that we commit an external library to VCS.
   * If I commit a file or place sub-modules directly to `/webextensions/`, I may use _vendor_.
* _third_party_ is more mozilla way, but I hesitate to use it for this project which is purely JS project.
* With a JavaScipt programmer hat, in today's JavaScript project,  I use npm to place an external libraries and it's rare case to vendor them without npm or other package system.
* Submodules are placed in _extlib_. So I seem _external_ or _lib_ might be suitable. I also think that using _extlib_ in `/webextensions/` would be confusable.

I don't have any strong opinion for that. If there is no other reasonable opinion, you can choose a like one.

piroor:
Thanks. Hmm, then I think I should use `extlib` for the purpose (because it clearly indicates that there are "libraries" from "external repository") and move current submodules to another place like `submodules`. I'll do that after merging this.

PR #1948, opened 2018-06-13
Use ES6 Collections for EventListenerManager's internals

saneyuki:
By this change, we can make it more expressive code.


piroor:
Thanks!

PR #1958, opened 2018-06-18
Enable no-var ESLint rule.

saneyuki:
* https://eslint.org/docs/rules/no-var
* Basically, I fix this change by `--fix` option of ESLint.
  But in some part, I also fix them by hand.


piroor:
Thanks!

PR #2005, opened 2018-08-21
add the actual shortcut key to the readme

LeoDJ:
It took me way too long to find out, which shortcut it was, that was being hard-coded. Please add it to the readme, so others don't have to struggle that much ;)


BobbyWibowo:
Ah, I was wondering why F1 didn't work, turns out it's no longer a preset and I had to manually add it to the new keyboard shortcuts option.

PR #2008, opened 2018-08-24
Properly handle mouse events when leaving or entering the sidebar

### LINKED ISSUE #1992: tab-mouseout event fired while still on tab
## Short description

27747f6fc433f8f452fed85f7cf0d29c5af5caf4 implemented support for `mouseover` and `mouseout` events. However, `mouseout` is currently not very useful, as it also fires when moving from a tab (`li`) to one of its children (`span`, e.g. the title or favicon). This makes it impossible to know when the cursor actually leaves the tab.

## Steps to reproduce

1. <details>
    <summary>Install the following example add-on:</summary>

    ```js
    const TST_ID = "treestyletab@piro.sakura.ne.jp";

    async function registerToTST() {
        try {
            var success = await browser.runtime.sendMessage(TST_ID, {
                type: "register-self",
                listeningTypes: ["tab-mouseover", "tab-mouseout"]
            });
        } catch(e) {}
    }

    browser.runtime.onMessageExternal.addListener(async (message, sender) => {
        if(sender.id !== TST_ID) { return; }
        let tab = message.tab;
        switch (message.type) {
            case "ready":
                registerToTST();
                break;
            case "tab-mouseover":
                console.log("enter tab " + tab.id);
                break;
            case "tab-mouseout":
                console.log("leave tab " + tab.id);
                break;
        }
    });
    registerToTST();
    ```
    *Note:* For me `mouseover` and `mouseout` only work after reopening the sidebar (`mousemove` works directly after installing the add-on). However, I haven't checked the dev-version or a clean profile, I just wanted to mention it in case you want to test the example.
    </details>
2. Move mouse over tab
3. Move mouse away from tab

## Expected result

```
enter tab 1
* mouse now on tab (title) *
leave tab 1
* mouse no longer on tab *
```

## Actual result

```
enter tab 1
leave tab 1
* mouse now on tab (title) *
* mouse no longer on tab (no event!) *
```

## Fixes

One solution is to ignore the `mouseout` event when the cursor leaves the tab for one of its children using the `relatedTarget` attribute of the event, which contains the target of the corresponding `mouseover` event using something like `!target.classList.contains("tab") || target.contains(e.relatedTarget)`. I think this also makes the current logic that prevents multiple consecutive events from the same tab obsolete.

Another solution is to use `mouseenter` and `mouseleave` events, which are [not firing](https://developer.mozilla.org/en-US/docs/Web/Events/mouseenter) when entering or leaving children of an element. However, these would have to be added to every tab individually, as they don't bubble.

I can create a pull request with any of these changes if you want.

<details>
<summary>Simple test of the first solution</summary>

```html
<nav id="tabs">
    <div class="tab" data-id="1" style="padding: 30px; background: orange;">
        <p style="height: 100px; background: gray;"></p>
    </div>
    <div class="tab" data-id="2" style="margin: 10px 0; padding: 30px; background: orange;">
        <p style="height: 100px; background: gray;"></p>
    </div>
    <div class="tab" data-id="3" style="padding: 30px; background: orange;">
        <p style="height: 100px; background: gray;"></p>
    </div>
</nav>
<script>
    let tabs = document.getElementById("tabs");

    tabs.addEventListener("mouseover", (e) => {
        // Ignore if not a tab or entering from a child (relatedTarget)
        if(!e.target.classList.contains("tab") || e.target.contains(e.relatedTarget)) {
            return;
        }
        console.log("enter: " + e.target.dataset.id);
    });
    tabs.addEventListener("mouseout", (e) => {
        // Ignore if not a tab or leaving to a child (relatedTarget)
        if(!e.target.classList.contains("tab") || e.target.contains(e.relatedTarget)) {
            return;
        }
        console.log("leave: " + e.target.dataset.id);
    });
</script>
```
</details>

## Environment

 * Platform (OS): Linux
 * Version of Firefox: 62b12
 * Version of Tree Style Tab: 2.4.24

##

klemens:
This new implementation for mouseover and mouseout is based on the relatedTarget property of the mouse event. It can be used to detect if an element is entered from a child, parent or from outside the window. The same is true for leaving elements.

This means we no longer have to keep track which element was entered or left last and directly send events without a timeout.

This mostly fixes #1992.

### DIFF
--- a/webextensions/sidebar/mouse-event-listener.js
+++ b/webextensions/sidebar/mouse-event-listener.js
@@ -164,22 +164,14 @@ onMouseMove = EventUtils.wrapWithErrorHandler(onMouseMove);
 
 function onMouseOver(event) {
   const tab = EventUtils.getTabFromEvent(event);
-  if (!tab || onMouseOver.lastTarget != tab.id) {
-    const lastMouseOutTab = Tabs.getTabById(onMouseOut.lastTarget);
-    if (lastMouseOutTab) {
-      TSTAPI.sendMessage({
-        type:     TSTAPI.kNOTIFY_TAB_MOUSEOUT,
-        tab:      TSTAPI.serializeTab(lastMouseOutTab),
-        window:   mTargetWindow,
-        ctrlKey:  event.ctrlKey,
-        shiftKey: event.shiftKey,
-        altKey:   event.altKey,
-        metaKey:  event.metaKey,
-        dragging: DragAndDrop.isCapturingForDragging()
-      });
-    }
-  }
-  if (tab && onMouseOver.lastTarget != tab.id) {
+
+  // We enter the tab element itself, but not from any of its children
+  const enterTabFromAncestor = event.target.classList.contains('tab') && !event.target.contains(event.relatedTarget);
+  // We enter the tab or any of its children from outside of the sidebar,
+  // which causes the relatedTarget (target of the mouseout event) to be null
+  const enterTabAndSidebar = tab && event.relatedTarget === null;
+
+  if (enterTabFromAncestor || enterTabAndSidebar) {
     TSTAPI.sendMessage({
       type:     TSTAPI.kNOTIFY_TAB_MOUSEOVER,
       tab:      TSTAPI.serializeTab(tab),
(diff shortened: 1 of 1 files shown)


piroor:
Thanks! I didn't know how to use `event.relatedTarget`.
https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/relatedTarget

PR #2010, opened 2018-08-27
Fix mouse events when moving over the close buttons

klemens:
Sorry, I missed one case in the previous pull request. :sweat_smile: Now everything works pretty well, as long as you don't move your mouse insanely fast.

> Moving from the close button of one tab directly to the close button of another tab without going through the tab element itself was not considered in 91f5d32e78. This commit fixes this by checking if the tab of the related target of the event is another valid tab.


piroor:
Thanks again!

PR #2011, opened 2018-08-28
Request optional permissions without toolbar button

Lej77:
[Firefox bug 1382953](https://bugzilla.mozilla.org/show_bug.cgi?id=1382953) has been fixed for Firefox 61 and Firefox ESR 60. This means that we can request optional permissions without using the toolbar button. 

This implementation attempts to request the permission without the toolbar button and if that fails it then tries to use the toolbar button instead. This can be useful since it allows permission to be request at times when the user hasn't done one of the interactions required to be allowed to request permissions. 

It also tracks if a request is currently pending and cancels all request until the current one is either allowed or denied. If this is not done then the user can click multiple times on the request button and the requests will be queued so that when the current request is denied it will appear again and again until the queue is emptied.

### DIFF
--- a/webextensions/common/common.js
+++ b/webextensions/common/common.js
@@ -168,6 +168,7 @@ export const configs = new Configs({
   simulateSVGContextFill: true,
 
   requestingPermissions: null,
+  requestingPermissionsNatively: null,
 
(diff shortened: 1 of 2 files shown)


piroor:
Thanks a lot!!

PR #2020, opened 2018-09-02
Generalize mouse move events to cover all possible cases

klemens:
I found another mouse move case that was not covered: Moving the mouse from an empty part of the tab bar directly to the close button. Instead of adding yet another special case, I generalized the first check and it suddenly included all other cases as well (including the new one I found). :smiley:

> Previously, the enter and leave events for a tab were derived from the (bubbling) mouseover and mouseout events by checking a number of different cases, like when entering the top level tab element, but not from any of its children or entering the tab directly without going through another element. However, it turned out that moving the mouse directly between child elements of two tabs was not covered, so another case was added in a0653c03f14ba2e9ad42162723e89dedd29c42c2.
> 
> Later, another missing case was discovered: moving the mouse from an empty part of the tab to directly onto a child element of a tab.
> 
> Instead of adding another special case, this commit simplifies all previously mentioned cases into a single, most general check: We emit a tab enter event, when the tab or any of its children is entered, but
> not from any other of the same tab's children. This also includes the case where we cannot determine where the mouse came from (eg. from outside the sidebar).

### DIFF
--- a/webextensions/sidebar/mouse-event-listener.js
+++ b/webextensions/sidebar/mouse-event-listener.js
@@ -164,18 +164,15 @@ onMouseMove = EventUtils.wrapWithErrorHandler(onMouseMove);
 
 function onMouseOver(event) {
   const tab = EventUtils.getTabFromEvent(event);
-  const relatedTab = Tabs.getTabFromChild(event.relatedTarget);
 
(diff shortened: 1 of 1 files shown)


piroor:
Thanks!

Issue #2006, opened 2018-08-23
Minor grammatical correction for group pages


ssokolow:
## Short description

The checkbox label "Close this tab when there is no more child" in group pages should be written as "Close this tab when there are no more children".

The singular "is no more" construction is only valid for countable nouns in English as an archaic/poetic construction with an inappropriate degree of emotional weight more similar to "is dead" or "has been destroyed" and, even if it were appropriate in this situation, it would still be written as "Close this tab when the last child is no more".

(On the other hand, "when there is no more water" would be valid, because "water" is an uncountable noun. You can have "no water" and "some water" but "a water" and "two waters" are only valid as slang for "a glass/cup/etc. of water" and "two glasses/cups/etc. of water")

## Environment

 * Platform (OS): Kubuntu 14.04.5 LTS
 * Version of Firefox: 62.0b19 (64-bit)
 * Version (or revision) of Tree Style Tab: 2.5.2

piroor:
Thanks, I've fixed that by fb0bece.

If you find more other mistakes, you can create pull requests easily by clicking the pencil button ("Edit") at https://github.com/piroor/treestyletab/blob/master/webextensions/_locales/en/messages.json

tushar-suresh:
This needed one more fix. Here's the PR - https://github.com/piroor/treestyletab/pull/2030

ssokolow:
Sorry that I got side-tracked on this. I agree that "child" needs to be "children" and I'll try to find some time to go through `messages.json` in the next few weeks.

(I've already started noting down some fixes, but I didn't have time to go through the entire file and some of the problems I did notice will take more time to solve, because I need to track down the context in which the message appears in order to know the proper correction.)

PR #2030, opened 2018-09-10
Fix minor typo in parent tab message

tushar-suresh:
Fixes https://github.com/piroor/treestyletab/issues/2006 .

Use plural form of child.


piroor:
Thanks!

PR #2041, opened 2018-09-28
Faster synchronization for loading throbbers

Lej77:
Sometimes a stutter is quite noticeable when the loading throbbers are synchronized. This mostly happens when I have many open tabs or many loading tabs. The stutter is because all throbbers are stopped and then re-started to synchronize them. This is supposed to happen when the animation has just completed so that it is less noticeable but when Firefox is under heavy load this can be delayed and cause the animation to stop halfway through and the re-start from the beginning. 

This pull request tries to improve on this in two ways:
* Less time between when animation is completed and the throbbers are re-started. This is done by triggering a reflow instead of waiting for it. The method used is from https://css-tricks.com/restart-css-animation/#article-header-id-0.
* A `#sync-throbber` is used for the timing of the throbber animation. This animation is not re-started when the sync logic is run. Note that the `#master-throbber` is still re-started so that unsynchronized throbbers that use it remain in sync with the synchronized throbbers. If there is a delay in the sync code then the `#sync-throbber`'s animation will be slightly ahead of the synchronized throbbers' animations. This means that the sync code will be triggered slightly before the synchronized throbbers complete their animations and they will hopefully be stopped just as they complete their animations. 

From my testing both of these fixes seems to work really well. Both together and separately. I did my tested in two different windows. One with 100 loading tabs and the other with 10. The loading tabs were not loading a website and continued loading forever. 

This pull request also fixes a probable bug in the `sidebar.html` file. The `#master-throbber-container` is not located inside the `favicon` element as it should be. This can be seen by setting the `#dummy-tabs` area's `z-index` to something greater such as ´500´. This was fixed by removing some new lines before and after the `#master-throbber-container`.

### DIFF
--- a/webextensions/sidebar/sidebar.html
+++ b/webextensions/sidebar/sidebar.html
@@ -60,6 +60,7 @@
     <ul id="dummy-tabs">
       <span id="dummy-favicon-size-box"></span>
       <span id="dummy-highlight-color-box"></span>
+      <span class="throbber"><span id="sync-throbber"></span></span>
       <li id="dummy-tab" class="tab">
         <span class="favicon"><span class="throbber" id="master-throbber-container"><span id="master-throbber"></span></span></span>
         <span class="label">-</span>
--- a/webextensions/sidebar/sidebar.html
+++ b/webextensions/sidebar/sidebar.html
@@ -61,9 +61,7 @@
       <span id="dummy-favicon-size-box"></span>
       <span id="dummy-highlight-color-box"></span>
       <li id="dummy-tab" class="tab">
-        <span class="favicon">
-          <span class="throbber" id="master-throbber-container"><span id="master-throbber"></span></span>
-        </span>
+        <span class="favicon"><span class="throbber" id="master-throbber-container"><span id="master-throbber"></span></span></span>
         <span class="label">-</span>
         <span class="counter">0</span>
         <span class="closebox"></button>
--- a/webextensions/sidebar/sidebar-tabs.js
+++ b/webextensions/sidebar/sidebar-tabs.js
@@ -27,7 +27,7 @@ let mInitialized = false;
 
 export function init() {
   mInitialized = true;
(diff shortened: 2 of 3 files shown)


webextensions/sidebar/sidebar.html
```diff
@@ -60,10 +60,9 @@
     <ul id="dummy-tabs">
       <span id="dummy-favicon-size-box"></span>
       <span id="dummy-highlight-color-box"></span>
+      <span class="throbber"><span id="sync-throbber"></span></span>
       <li id="dummy-tab" class="tab">
-        <span class="favicon">
-          <span class="throbber" id="master-throbber-container"><span id="master-throbber"></span></span>
-        </span>
+        <span class="favicon"><span class="throbber" id="master-throbber-container"><span id="master-throbber"></span></span></span>
```
piroor: This change looks unnecessary. Could you restore old indentation?

piroor:
Thanks! This looks good, so I'm doogfooding now. Detailed descriptions also helps me to understand what happens.

Lej77:
@piror It actually changes the layout of the hidden master throbber though. As I mentioned at the end of my first comment. With the old code Firefox adds text elements before and after the master throbber container in the dummy tab's favicon element. The spaces seems to be interpreted as actual new line text elements. To see this yourself just ensure that the dummy tabs area is visible by setting it's z index to something greater and then reload a tab to see where the hidden throbber is. It will be below the dummy tab.

Lej77:
@piror It actually changes the layout of the hidden master throbber though. As I mentioned at the end of my first comment. With the old code Firefox adds text elements before and after the master throbber container in the dummy tab's favicon element. The spaces seems to be interpreted as actual new line text elements. To see this yourself just ensure that the dummy tabs area is visible by setting it's z index to something greater and then reload a tab to see where the hidden throbber is. It will be below the dummy tab.

PR #2082, opened 2018-12-01
Unprefix -moz-linear-gradient()

chris-morgan:
As of https://bugzilla.mozilla.org/show_bug.cgi?id=1337655, Nightly
doesn’t support the obsolete -moz-linear-gradient() syntax; and the plan
is to drop it on stable eventually too.

This involves the inversion in direction specification between the old
and current syntaxes: “left” becomes “to right”, &c. The default
side-or-corner is “to bottom”, but I decided to leave those ones in
rather than dropping it. Clarity, and all that: not everyone knows the
default direction of a gradient.


piroor:
Wow, thanks!

PR #2083, opened 2018-12-02
Locale: Shorter message for animations config

ratijas:
I mean, keep it simple. It already a checkbox, so why "enable"/"disable" text? And what's with those "effects" anyway — aren't they just "animations"?


ratijas:
Whatever.  Just sounds dumb and redundant for me.

PS Firefox is not the truth of last resort when it comes to UI/UX (imho, of course).

PR #2091, opened 2018-12-14
Use filter for browser.tabs.onUpdated event.

Lej77:
[Firefox 61 and later supports using a filter](https://bugzilla.mozilla.org/show_bug.cgi?id=1329507) for the [browser.tabs.onUpdated event](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/onUpdated). This should improve performance. 

I have tested this code change for TST v2.6.8 and it works for Firefox 60 (ESR) and Firefox 64.0 (stable).

### DIFF
--- a/webextensions/common/api-tabs-listener.js
+++ b/webextensions/common/api-tabs-listener.js
@@ -61,8 +61,20 @@ function logUpdated(...args) {
 }
 
 export function startListen() {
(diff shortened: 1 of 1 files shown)


piroor:
Oh, thanks a lot!

PR #2092, opened 2018-12-14
 Don't populate window returned by browser.tabs.highlight. 

Lej77:
By default the [browser.tabs.highlight](https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/tabs/highlight) function returns the populated window that was changed. But we can instead return an unpopulated window for better performance if we aren't using the returned value (which we aren't). I have not tested this changes but according to the API reference this property is support in Firefox 63 the same as the function so there shouldn't be any issues.

More info: [Firefox bug 1489814](https://bugzilla.mozilla.org/show_bug.cgi?id=1489814)


piroor:
Thanks!

PR #2237, opened 2019-04-22
feat(focus): Add more shortcuts to focus the child/sibling

15cm:
I strongly feel current keyboard shortcuts are not enough for users, e.g. me, to navigate the TST with keyboard. 3 shortcuts are added in this PR. I would like to explain why they are useful here:

1. Move focus to the last child
When we config TST to open the link as a new child, the child will be appended to the end of the current tree. Intuitively we will try to reach the last created child that is at the end. But these was no a shortcut to do that.

2. Move focus to the previous/next sibling
Currently TST has 3 ways to navigate among tabs: a level up(+1 level, focus parent), a level down(-1 level, focus child) and level irrelevant(focus previous/next). Lack of shortcuts to focus siblings(0 level) makes it painful to navigate between deep trees in the same level with keyboard.
With these two shortcuts, we can make most of the tab navigation by combining the +1, 0, -1 level movement with keyboard.

Hence I believe these shortcuts are useful tools that should not burdens the users.


piroor:
Thanks!

Issue #2238, opened 2019-04-22
v3.0.7 Critically Broken: 1st Tab Never Loads, Last Never Clickable, Wrong Display Order, Clicking Temporarily Activates Wrong Tab, syncTabsOrder() Never Completes, Need Auto Reinit on Mismatched Tabs vs. Thrown Errors


PowerAccess:
@piroor 

The latest TST v3.0.7 (obtained from Git) has numerous critical errors rendering is essentially unusable, from the moment first opened, which I've summarized below, together with associated logged errors and debugging results.

### Related Issues Summarized Here

I've also summarized here (and in my following comments/replies here) the most important points, debugging results and potential fixes from my most recent replies to related issue #2199, which should be easier to follow than that longer thread (though you can also find further details there if needed).

### Summary of New Critical Issues with Latest Version (v3.0.7 from Git)

1. **Tabs Displayed in Wrong Order**
   - In the newest of the recent critical issues, Tab display order is now often wrong
   - For example, in the below screenshot, "Amazon.com - Your Subscribe & Save" tab shown is shown at the bottom of TST, when it should be the 2nd tab (after "Home")
   - Also, when I try to click or close the "Amazon.com - Your Subscribe & Save" tab (shown incorrectly at bottom) in TST sidebar, nothing happens.
   - And, other tabs opened later on (as seen in other screenshots further below) also appear out of order, so this isn't just limited to the last tab being wrong.

![image](https://user-images.githubusercontent.com/4053734/56533575-c758a880-650c-11e9-818f-a9a2483c6624.png)

2. **First Tab Never Shown as Loaded**
   - Also, the first tab (Walgreens. Trusted) just shows the animated loading favicon and fails to even show the page title ("Walgreens. Trusted" seen in tabbar) and instead just shows the URL, as if it never loaded.
   - See follow-up comment here for detailed Debugging Results and Steps to Reproduce.
   - Notice how the tab had in fact fully loaded (10 minutes before even) without issue as seen in tabbar in this screenshot
   - This issue seems to occur for any tab(s) opened while TST is still initializing (while Sidebar is opening, but before populated with tabs and/or possibly for tabs opened before sidebar even auto-opens on startup)
   - Many "Unchecked lastError" errors also appear in Console immediately before the "Could not establish connecction" error, and, according to discussions I've seen elsewhere for Firefox, all those occur when attempting to send a message when no listener is yet registered.
[...]

PR #2239, opened 2019-04-23
Fixed syncTabsOrder() / Unclickable Tabs (#2199) + Many (Not all) of v3.0.7 Critical Issues (#2238) + Auto Reload Sidebar on Critical Tab Order Failures

PowerAccess:
### Fixes for #2199 (with syncTabsOrder) + many (but not all) of #2238 (v3.0.7 Critical Issues), including:
- Fixed syncTabsOrder() never actually working (ever) due to always either deferring indefinitely or failing the initial conditions or comparing the wrong things, with a complete rewrite of it
- Added support for automatic complete rebuild of trees when syncTabsOrder() fails
- Added "Reload Sidebar" to context menu for sidebar, which users can use when encounter other critical issues like with unclickable tabs, or in wrong order, etc.
- Fixed "Error: Could not establish connection. Receiving end does not exist" at least when either:
   A. caching is disabled (configs.useCachedTree = false) OR 
   B. if configs.useCachedTreeBackgroundExport = false (left = true for now) (not exposed in Options UI)

### Remaining Issues from #2238 still not fixed

1. "Error: Could not establish connection. Receiving end does not exist" still occurs when caching is enabled
   - I added configs.useCachedTreeBackgroundExport option 
   - I left true by default for now 
   - I suggest you set = false after testing further with session restore cache use (may prevent that, or that may be already be prevented due to that error remaining anyways)
2. Tabs opened soon after Firefox startup (with TST sidebar auto shown on startup) *while TST is initializing*
   - For example if was to open "https://trakt.tv/calendars/my/shows", end up with animated loading icon showing forever,
   - Sometimes (not always) also shows only URL (instead of title) and/or has FaviconLoader error 
3. sendMessage(kCOMMAND_PULL_TABS) in rebuildAll() in background.js is returning undefined
   - At least this occurs when this is auto-called early on during init, but maybe anyways before too
   - Not mentioned in #2238 yet, but may be related to above 2 remaining issues
   - I fixed the error to handle undefined, but may be underlying issue causing this to happen which may relate to remaining unfixed startup issues described in #2238

### DIFF
--- a/webextensions/_locales/en/messages.json
+++ b/webextensions/_locales/en/messages.json
@@ -17,6 +17,7 @@
   "command_focusNextSilently":     { "message": "Focus to Next Tab (don't expand tree)" },
   "command_focusParent":           { "message": "Focus to Parent Tab" },
   "command_focusFirstChild":       { "message": "Focus to First Child Tab" },
+  "command_reloadSidebars":        { "message": "Reload Sidebar" },
   "command_tabbarUp":              { "message": "Scroll Tabs Up by Line" },
   "command_tabbarPageUp":          { "message": "Scroll Tabs Up by Page" },
   "command_tabbarHome":            { "message": "Scroll Tabs to Top" },
--- a/webextensions/background/background.js
+++ b/webextensions/background/background.js
@@ -113,6 +113,7 @@ export async function init() {
   onBuilt.dispatch();
   MetricsData.add('init: started listening');
 
+  //provide reloadSidebars() implementation for use in Tab context menu (avoiding cyclical dependencies)
   const reloadSidebarsCommand = reloadSidebars;
   TabContextMenu.init(reloadSidebarsCommand);
   MetricsData.add('init: started initializing of context menu');
@@ -120,13 +121,20 @@ export async function init() {
   Permissions.clearRequest();
 
   for (const windowId of restoredFromCache.keys()) {
+
+    //MAYBE: Is await needed for lines commented with "//await needed?"
+    //Can remove some of the 4 added await uses, possibly later, to speed-up,
+    //but suggest doing *after* fixing the remaining major initialization issues from #2238 (in case this help to avoid or pinpoint them) like:
+    // "Error: Could not establish connection. Receiving end does not exist" 
+    // and "Tab opened during init never sets favicon or sometimes title"?
+  
     if (!restoredFromCache[windowId])
-      await BackgroundCache.reserveToCacheTree(windowId);
-    await TabsUpdate.completeLoadingTabs(windowId);
+      await BackgroundCache.reserveToCacheTree(windowId); //await needed?
+    await TabsUpdate.completeLoadingTabs(windowId); //await needed?
   }
 
   for (const tab of Tab.getAllTabs(null, { iterator: true })) {
-    await updateSubtreeCollapsed(tab);
+    await updateSubtreeCollapsed(tab); //await needed?
   }
   for (const tab of Tab.getActiveTabs()) {
     for (const ancestor of tab.$TST.ancestors) {
@@ -139,7 +147,7 @@ export async function init() {
 
   // we don't need to await that for the initialization of TST itself.
   //waiting to see if helps fix the m
-  await MetricsData.addAsync('init: initializing API for other addons', TSTAPI.initAsBackend());
+  await MetricsData.addAsync('init: initializing API for other addons', TSTAPI.initAsBackend()); //await needed?
 
   mInitialized = true;
   onReady.dispatch();
--- a/webextensions/background/background.js
+++ b/webextensions/background/background.js
@@ -113,19 +113,20 @@ export async function init() {
   onBuilt.dispatch();
   MetricsData.add('init: started listening');
 
-  TabContextMenu.init();
+  const reloadSidebarsCommand = reloadSidebars;
+  TabContextMenu.init(reloadSidebarsCommand);
   MetricsData.add('init: started initializing of context menu');
 
   Permissions.clearRequest();
 
   for (const windowId of restoredFromCache.keys()) {
     if (!restoredFromCache[windowId])
-      BackgroundCache.reserveToCacheTree(windowId);
-    TabsUpdate.completeLoadingTabs(windowId);
+      await BackgroundCache.reserveToCacheTree(windowId);
+    await TabsUpdate.completeLoadingTabs(windowId);
   }
 
   for (const tab of Tab.getAllTabs(null, { iterator: true })) {
-    updateSubtreeCollapsed(tab);
+    await updateSubtreeCollapsed(tab);
   }
   for (const tab of Tab.getActiveTabs()) {
     for (const ancestor of tab.$TST.ancestors) {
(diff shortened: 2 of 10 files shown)


piroor:
On the branch https://github.com/piroor/treestyletab/tree/PowerAccess-master I'm trying to understand changes and reformat them to match to the design policy of TST's existing codes...

piroor:
@PowerAccess I've done reformatting of this PR at the branch https://github.com/piroor/treestyletab/tree/PowerAccess-master . Does it still work effectively as you expected? Did I misunderstood something?

piroor:
Sorry I'm still confusing. This PR looks to having too much things, and they looks should be separated to different changes - especially reducing of the "Error: Could not establish connection. Receiving end does not exist" warning. I think we need to think "waiting until completion to do things stability" and "waiting until completion to reduce warnings" separately. You added many "await" around the initialization process, but I thought that their purpose may be confused. How do you think about that point?

piroor:
After a research, I've found an information about the warning: the "Could not establish connection" warning appears when `runtime.sendMessage()` is called before any listener is added to the `runtime.onMessage`. If the description is correct, I think I need to reduce the priority for suppressing of that, when such suppressing doesn't increase stability.

piroor:
Now I'm trying to import changes introduced with this PR to the master branch step by step.

piroor:
Based on the idea of this PR, I've added codes to rebuild all at both the master process and sidebars, when the list of tabs tracked by the master process is not matched to the native tabs. @PowerAccess, could you confirm that these changes are effective or not to solve "failed to sync" cases on your environment?

PowerAccess:
Thanks for integrating some of those fixes from my PR, and continuing with that.
After your changes, some of the issues I had fixed remain fixed or improved, but others remain broken.

Tab sync issues still are occurring, though less often. Now, with latest in Git master, when a tab fails to sync and becomes "unclickable" it doesn't cause all the other tabs created after that to also become unclickable.

However, tabs still get out-of-sync and remain out-of-sync forever, even 10 minutes later.
The whole implementation of syncTabsOrder() was incorrect I'd found when debugging.  It never actually corrected anything (for the few times it was called, which was very rare in the first place). 

I explained why in past issue replies in more detail a few of the problems encountered with it, and then encountered even more, with the basic design of it even. That is why I ended up rewriting it, and would suggest you use the version I had written it to.

Can you use at least my implementation of syncTabsOrder() as-is vs. applying just parts of it?

Also, can you have syncTabsOrder() called automatically every configurable interval (defaulting to 10 seconds ideally, or at least 30 or 60 sec)?

Even my rebuildSidebars() implementation ended up with executing so fast and without even any kind of flicker (since it did a lightweight resync more then reload) so that there was no noticeable delay or performance overhead, so I think frequent calls to this should work well. Also, because syncTabsOrder() was almost never called (even when I changed reserveSyncTabsOrder() to execute it immediately, as I suggest you also do), it's even more critical to do that periodically. 
 
Plus, I would suggest providing Reload Sidebar like I had implemented, to allow users to recover from more rare issues, to workaround remaining issues (like still have now) and as a workaround for if/when syncing gets broken (as I've noticed more and more tab sync issues occur over time, and can easily get broken again).

PowerAccess:
### Errors Occurring for Remaining Unclickable Tabs Issue

Here are some of the errors I'm encountering with _latest from Git master_, when I found some tabs unclickable. 

For example, in this case, when I try to click on "v3.0.7" tab (which is the parent of the active tab in this screenshot below) then it activates the tab I clicked on (at least, which is better than a random tab like before) for like 100ms before auto-switching back to whatever the previous tab was.


image

### Errors in Console with Latest Git Master

15:24:39.745 Error: Incorrect argument types for tabs.query. 2 sidebar-tabs.js:408:18
15:29:52.626 TypeError: "tab is null"
    onClick moz-extension://d9f19977-38ac-4464-a14f-a0309441d127/background/context-menu.js:257
    results moz-extension://d9f19977-38ac-4464-a14f-a0309441d127/extlib/EventListenerManager.js:61
    dispatchWithDetails moz-extension://d9f19977-38ac-4464-a14f-a0309441d127/extlib/EventListenerManager.js:54
    dispatch moz-extension://d9f19977-38ac-4464-a14f-a0309441d127/extlib/EventListenerManager.js:43
    onMessage moz-extension://d9f19977-38ac-4464-a14f-a0309441d127/background/tab-context-menu.js:759
EventListenerManager.js:87:17
15:29:52.633 TypeError: result is undefined
EventListenerManager.js:102:11
15:30:10.028 Error: Incorrect argument types for tabs.query. 2 sidebar-tabs.js:408:18

PowerAccess:
### Tabs Opened During TST Init Issues

Also, the Remaining Issue with "Tabs Opened Early During TST Init Always having Loading Animated Favicon" still remains, as detailed in [my last reply](https://github.com/piroor/treestyletab/issues/2238#issuecomment-485681869) to Issue #2238.  

This might apply to tabs opened before TST sidebar fully initializes too, but I tested with TST sidebar always auto opening on Firefox startup so it least occurs for any tabs you open while TST sidebar is shown but still initializing (eg. with progress bar shown).

See Steps to Reproduce provided and try using a bookmark in Bookmark Toolbar for "https://trakt.tv/calendars/my/shows" specifically (in case easier to reproduce with that) so that you can click on it to open it very quickly before TST finishes initializing.

This is part of why I spent time on the errors like "`Could not establish connection`" and waiting for complete initialization, because tabs opened early during TST init end up corrupted, so that even Reload Sidebar would not fix them (though Reload Tab did).

Please note that my PR did not fix (or at least not completely fix) this issue, though I made some changes which I hoped would help narrow down possible causes of the issue or avoid other issues with tabs opening tabs during TST init, since those tabs always have major issues with them currently, that even Reload Sidebar can't fix.

One issue I found which may be relevant to that, when debugging latest from Git master, is:
"SecurityError: The operation is insecure." occurs at
`data = this.canvas.toDataURL('image/png');`

```
at onLoad() (TabFavIconHelper.js line 207) 
called by getEffectiveUrl() (TabFavIconHelper.js line 281)
```

If you enable "Pause on Exceptions" and follow Steps to Reproduce from linked to reply, then you may be able to see that as well:
image

piroor:
Thank you for comments. I've not read all yet, but I think we need to build a consensus about the basic strategy of this project.

* Basically I'm negative to add a context menu command "Reload Sidebars". It will help development actually, but I think that it shouldn't be exposed to users by default. For example I rejected to adding "open TST options" command for a long time, because I have a policy; if people need to open options page too frequently, something wrong. Less accessibiltiy to the options page is not the problem we should resolve, instead, we need to struggle to reduce need to change options , with implementing of intelligent behaviors and other efforts.
* I agree that periodical sync will increase stability. But I need to investigate about other solutions before activating it by default. Periodical sync is a supportive care, but something definitive therapy is better than that. Supportive care enabled by default may reduce motivation to solve the problem fundamentally. (For example I didn't try to optimize the initialization process for a long time because the cache system worked well for me.)

From now I try to read your comments to understand things you actually did...

piroor:
Oops, sorry there were some careless mistakes as you commented at https://github.com/piroor/treestyletab/pull/2239#issuecomment-486005190 , and I've fixed them.

piroor:
After reading the comment https://github.com/piroor/treestyletab/issues/2238#issuecomment-485681869 I've tried the step "Once sidebar is shown, but while still shows loading progress (before fully initialized) I immediately click on a bookmark for:" and I successfully reproduce a failure case: some tabs are not tracked by the master process. Those tabs are opened at the timing between `browser.windows.getAll()` and `ApiTabsListener.startListen()`:

```javascript
// background/background.js
    waitUntilCompletelyRestored().then(() => {
      // don't wait at here for better performance
      promisedWindows = browser.windows.getAll({
        populate:    true,
        windowTypes: ['normal']
      }).catch(ApiTabs.createErrorHandler());
    }),
  ...
  // here!
  ...
  ApiTabsListener.startListen();
```

This problem looks unsolvable, because any fundamental solution may make the initialization process blocked and slow. Now I agree that "reload all" operation is required on such cases as a workaround, and I've introduced more fixes to do the auto-reload-all operation triggered by the function `syncTabsOrder()`.

piroor:
> One issue I found which may be relevant to that, when debugging latest from Git master, is:
"SecurityError: The operation is insecure." occurs at

After a research I've realized that it happens when I open a new tab `about:newtab`. The favicon of the page is `chrome://branding/content/icon32.png` and `toDataURL()` raises a security error on such cases. It should be suppressed with the commit https://github.com/piroor/webextensions-lib-tab-favicon-helper/commit/f92cedd28ebe1edcbbfaf4693a5ad32dd7396d75

piroor:
After more research I've realized that `TabsUpdate.completeLoadingTabs()` didn't worked as I expected. Tabs opened while TST is initializing itself will be completed their "loading" state correctly, with the commit 1e545670135314977789e05c92139d931ddc7592.

piroor:
> ### Remaining Issues from #2238 still not fixed
> 
> 1. "Error: Could not establish connection. Receiving end does not exist" still occurs when caching is enabled

As I commented at https://github.com/piroor/treestyletab/issues/2238#issuecomment-486172115 , I think this kind warning on initialization is a side effect of optimizations.

> 2. Tabs opened soon after Firefox startup (with TST sidebar auto shown on startup) *while TST is initializing*

I thought that it looked impossible to be "resolved" completely due to TST's internal design. After a reconsideration I've realized that those tabs can be tracked without auto-reload. The commit aaf9e11ea does that. Now all tab events notified while TST is building tree with tabs fetched via `windows.getAll()` are listened and stacked until initialization is finished. On my environment, without auto-reload, all tabs I opened while initialization are tracked as expected. Wow!

>    - For example if was to open "https://trakt.tv/calendars/my/shows", end up with animated loading icon showing forever,

This should be fixed with the commit 1e54567. ( https://github.com/piroor/treestyletab/pull/2239#issuecomment-486123741 )

>    - Sometimes (not always) also shows only URL (instead of title) and/or has FaviconLoader error 

I think that the error should be suppressed with some changes I already did. ( https://github.com/piroor/treestyletab/pull/2239#issuecomment-486117122 )

> 3. sendMessage(kCOMMAND_PULL_TABS) in rebuildAll() in background.js is returning undefined
>    - At least this occurs when this is auto-called early on during init, but maybe anyways before too
>    - Not mentioned in #2238 yet, but may be related to above 2 remaining issues
>    - I fixed the error to handle undefined, but may be underlying issue causing this to happen which may relate to remaining unfixed startup issues described in #2238

As I commented at https://github.com/piroor/treestyletab/issues/2238#issuecomment-486172115 , I think they are not bugs and just side effects from aggressive optimizations.

PowerAccess:
@piroor 

Thanks for your help. I've been testing versions from Git and then XPI releases for a while and seems like all major issues are now resolved in v3.0.11 XPI release.

PowerAccess:
@piroor 

Thanks for your help. I've been testing versions from Git and then XPI releases for a while, which now include much of my fixes from this PR, and seems like all major issues are now resolved in v3.0.11 XPI release for #2199 and #2238.

PR #2252, opened 2019-05-01
Complete German translation

SammysHP:
There might be a conflict in the hotkey mapping of the context menu and some new translations have no mapping yet. On the other hand the mappings were broken before, so this is not a regression.


piroor:
Thanks a lot!

Issue #2300, opened 2019-06-04
Addons that extend TST randomly stop working when restarting Firefox


xzn:
Firefox 67 on Windows 10 64-bit

Not sure how to reproduce this,

I have quite a few addons installed. Aside from TST and a bunch of other addons that I'm using, a few addons that integrate with TST are used:

TST Middle Click
TST Mouse Wheel
Tab Flip
Unload Tabs
Move Unloaded Tabs

Occasionally when I restart Firefox some of these addons stop working and I have to disable and re-enable them again in Addons Manager.

piroor:
I might reproduced this problem on my environment, but it is not consistently.

I still cannot find out the trigger, but when the problem happens, the ID of the addon which didn't work correctly is missing on the list of known addons which will receive `ready` notifiaction (`configs.cachedExternalAddons`). In this case, TST and the addon (for example, Tab Flip) will work as:

1. Tab Flip is loaded.
2. Tab Flip tries to send `register-self` to TST, but TST is not initialized yet.
3. TST is loaded.
4. TST tries to send a `ready` notification message to known helper addons. Then Tab Flip is not listed in the list of known addons, thus `ready` notification won't be sent to Tab Flip.
5. As the result, both TST and Tab Flip are installed and activated but they don't work together.

So the problem is: why Tab Flip is not listed in the list of known addons? When it is removed from the list? Or, why it was not added to the list on its installation?

piroor:
OK, I've realized whats happen on a failure case.

1. Other helper addons are loading. They try to send `register-self` message to TST. No one listens the message, so nothing happen.
2. Firefox starts loading of TST. `configs.cachedExternalAddons` is still blank.
3. TST starts listening of messages from other addons.
4. A helper addon sends a `register-self` message to TST.
5. TST receives the `register-self` and tries to register the helper addon. **Then `configs.cachedExternalAddons` is filled only with the ID of the helper addon, instead of last saved list of known addons.**
6. TST is completely loaded, and TST tries to send `ready` messages to known addons. But the list of known addons just contains the addon registered at the step 5. Others loaded before TST at the step 1 won't receive the `ready` notification.

piroor:
The commit 8e68b89 should fix this problem.

xzn:
I've just tested the latest commit 9584a3ea0d06545ed61e54f7023d52dc4326da9c on nightly and it appears to be fixed.

[...]

PR #2301, opened 2019-06-05
Load configs before registering onMessageExternal #2300

xzn:
See #2300


piroor:
Thanks! Hmm, but I think that the current code is safer than this version because TST's background page may not be initialized completely when `configs.$loaded` is resolved.

xzn:
Fair enough!

Closing.

PR #2306, opened 2019-06-07
Move 'tab-clicked' message response handling next to 'tab-mouseup'

xzn:
Match behavior described here:

https://github.com/piroor/treestyletab/wiki/API-for-other-addons#when-a-tab-is-clicked

Alternatively there's [this workaround](https://framagit.org/ariasuni/tst-tabflip/merge_requests/2) for tst-tabflip. Whichever gets merged first...


piroor:
Thanks! Hmm, but you looks don't need to import a background module `HandleTabMultiselect` from the sidebar, instead you just need to move `HandleTabMultiselect.updateSelectionByTabClick` to `webextensions/sidebar/mouse-event-listener.js`. And, there looks to be no reason to keep codes for `Constants.kNOTIFY_TAB_MOUSEDOWN` at `handle-misc.js` anymore. Moving those codes to `webextensions/sidebar/mouse-event-listener.js` together looks to make things more clear.

And one more concern: is this compatible to Multiple Tab Handler and other API clients?

xzn:
Additional changes per request:

Moved `Constants.kNOTIFY_TAB_MOUSEDOWN` related stuff from `handle-misc.js` to `mouse-event-listener.js`.
Moved `updateSelectionByTabClick` to `mouse-event-listener.js`, now named `updateMultiselectionByTabClick`.
Removed `Constants.kNOTIFY_TAB_MOUSEDOWN_CANCELED` related stuff since it's no longer used.

As for MTH there doesn't seems to be any incompatibility. The API should work just like before except that, if any one of `tab-mousedown`/`tab-mouseup`/`tab-clicked` is cancelled the all related tab operation will be cancelled. Whereas before `tab-mouseup` is handled separately from `tab-mousedown`/`tab-clicked`.

webextensions/background/handle-misc.js
```diff
@@ -26,15 +26,10 @@ import * as Background from './background.js';
 import * as TabsGroup from './tabs-group.js';
 import * as Tree from './tree.js';
 import * as Commands from './commands.js';
-import * as HandleTabMultiselect from './handle-tab-multiselect.js';
```
piroor: This module still need to be loaded from anywhere on the background page (for example background/index.js or background/background.js) to multiselect collapsed tree.

xzn: Sorry wasn't really understanding how javascript import works. In any case you appear to have already been loading this module in [`background/index.js`](https://github.com/piroor/treestyletab/commit/7d5a8b06d7adf894ba6d8fac0488c6c2f74532b0#diff-9f54d9eadb1d83ec226914ccb3adb65aR18). I just tested that multiselect collapsed tree (where collapsed tabs are automatically selected as well) still works.

piroor: Oh, you right. I missed that it was already loaded from index.js.

webextensions/sidebar/mouse-event-listener.js
```diff
@@ -256,12 +261,25 @@ function onMouseDown(event) {
...
-    windowId: mTargetWindow,
-    tabId:    tab && tab.id
-  })).catch(ApiTabs.createErrorHandler());
+  mousedown.promisedMousedownNotified = Promise.all([
+    browser.runtime.sendMessage({type: Constants.kNOTIFY_TAB_MOUSEDOWN })
+      .catch(ApiTabs.createErrorHandler()),
+
+    (async () => {
+      log('Constants.kNOTIFY_TAB_MOUSEDOWN');
+      await Tab.waitUntilTracked(mousedownDetail.tab);
```
piroor: This check (L270-L273) looks needless anymore. Originally this was introduced for safety about cross-namespace asynchronous operation, and now the `tab` (got with `const tab = EventUtils.getTabFromEvent(event) || EventUtils.getTabFromTabbarEvent(event);`) looks certainly available.

piroor:
Anyway, almost looks good to merge. Thanks a lot!

piroor:
After merging I've done some more changes. If my changes break functionalities, please comment here.

PR #2354, opened 2019-08-10
Add the container name to the classes on a tab.

Arthaey:


piroor:
I think you need to remove old class for the previous cookieStoreId.

Otherwise `setAttribute()` looks better than `addState()` for the purpose. Then you can use the attribute selector like: `.tab[container-name="firefox-default"] { ... }`

PR #2355, opened 2019-08-10
Add <span class="left-edge"> to each tab.

Arthaey:
It's otherwise hard to use CSS pseudo-elements to add things that
align on the far left of the sidebar, due to tab indenting.


piroor:
Could you describe some examples what really impossible is and what become available with this change?

piroor:
And, some screenshots may help me to understand the importance of this change.

PR #2368, opened 2019-08-22
[WIP] Allocate less

bb010g:
TST likes to suck up my CPU when it's doing nothing due to causing a bunch of tiny GC runs while mostly idle. Sometimes those pauses are noticeable in FX's UI. This PR helps by avoiding a ton of allocations on hot idle spots, but there are still significant issues while doing things like dragging tabs around or even opening new tabs (though to a lesser extent). Tested on a browser with ~1400 real tabs currently open, but without a lot of the more nested tree structure that normally makes this worse because somehow my session borked. (If you want me to profile more with that original worse-performing setup, I can do so by reorganizing my current session back to how it was; just say so and I'll try to speed up that when I've got time.)

The main thing is to stop using friendly APIs that copy a lot when dealing with things like the whole set of tabs, because that gets way more pathological as tab counts grow. JavaScript doesn't have Haskell's stream fusion. (Side thought: it might be helpful rewriting TST's core in something that emits Wasm, like AssemblyScript or Rust, so you can deal with the amortized cost of a linear memory slab and probably allocate less on top, reducing CPU pressure? I'm not sure if it'd be worthwhile, though; skimming the newest profile's allocation metrics makes me think heavier knowledge of the TST addon API & code would be required to move forward.)

This could be merged now, but I'm not sure about how to reduce some of the bigger allocation issues. (I'll upload some profiles below for examples on just switching tabs while otherwise idle, and actually idling. I unfortunately forgot to save my original profiles, but the GC pauses were a lot worse & more consistent.)

[idle-profile.json.gz](https://github.com/piroor/treestyletab/files/3532290/idle-profile.json.gz)

[idle-ish-profile.json.gz](https://github.com/piroor/treestyletab/files/3532297/idle-ish-profile.json.gz)

---
This change is [Reviewable](https://reviewable.io/reviews/piroor/treestyletab/2368)

### DIFF
--- a/webextensions/background/tree.js
+++ b/webextensions/background/tree.js
@@ -29,6 +29,7 @@
 import TabIdFixer from '/extlib/TabIdFixer.js';
 
 import {
+  filterMap,
   log as internalLogger,
   wait,
   dumpTab,
@@ -1075,8 +1076,7 @@ export async function moveTabs(tabs, options = {}) {
       const startTime = Date.now();
       const maxDelay = configs.maximumAcceptableDelayForTabDuplication;
       while (Date.now() - startTime < maxDelay) {
-        newTabs = movedTabs.map(tab => Tab.get(TabIdFixer.fixTab(tab).id));
-        newTabs = newTabs.filter(tab => !!tab);
+        newTabs = movedTabs.filterMap(Tab.get(TabIdFixer.fixTab(tab).id) || undefined);
         if (mSlowDuplication)
           UserOperationBlocker.setProgress(Math.round(newTabs.length / tabs.length * 50) + 50, windowId);
         if (newTabs.length < tabs.length) {
@@ -1115,8 +1115,7 @@ export async function moveTabs(tabs, options = {}) {
   }
 
 
-  movedTabs = movedTabs.map(tab => Tab.get(tab.id));
-  movedTabs = movedTabs.filter(tab => !!tab);
+  movedTabs = movedTabs.filterMap(tab => Tab.get(tab.id) || undefined);
   if (options.insertBefore) {
     await TabsMove.moveTabsBefore(
       movedTabs,
@@ -1136,8 +1135,7 @@ export async function moveTabs(tabs, options = {}) {
   }
   // Tabs can be removed while waiting, so we need to
   // refresh the array of tabs.
-  movedTabs = movedTabs.map(tab => Tab.get(tab.id));
-  movedTabs = movedTabs.filter(tab => !!tab);
+  movedTabs = movedTabs.filterMap(tab => Tab.get(tab.id) || undefined);
 
   return movedTabs;
 }
@@ -1176,9 +1174,8 @@ export async function openNewWindowFromTabs(tabs, options = {}) {
       log('moved tabs: ', movedTabs.map(dumpTab));
       const movedTabIds     = movedTabs.map(tab => tab.id);
       const allTabsInWindow = window.tabs.map(tab => TabIdFixer.fixTab(tab));
-      const removeTabs      = allTabsInWindow
-        .filter(tab => !movedTabIds.includes(tab.id))
-        .map(tab => Tab.get(tab.id));
+      const removeTabs      = filterMap(allTabsInWindow, tab =>
+        !movedTabIds.includes(tab.id) ? Tab.get(tab.id) : undefined);
       log('removing tabs: ', removeTabs.map(dumpTab));
       TabsInternalOperation.removeTabs(removeTabs);
       UserOperationBlocker.unblockIn(newWindow.id);
--- a/webextensions/common/common.js
+++ b/webextensions/common/common.js
@@ -305,7 +305,7 @@ export const configs = new Configs({
 
   testKey: 0 // for tests/utils.js
 }, {
-  localKeys: `
+  localKeys: filterMap(`
     optionsExpandedSections
     sidebarPosition
     sidebarDirection
@@ -324,7 +324,10 @@ export const configs = new Configs({
     requestingPermissions
     requestingPermissionsNatively
     testKey
-  `.trim().split('\n').map(key => key.trim()).filter(key => key && key.indexOf('//') != 0)
+  `.trim().split('\n'), key => {
+    key = key.trim();
+    return key.indexOf('//') != 0 ? key : undefined;
+  })
 });
 
 configs.$loaded.then(() => {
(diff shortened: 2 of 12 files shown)


webextensions/background/tree.js
```diff
@@ -1075,8 +1076,7 @@ export async function moveTabs(tabs, options = {}) {
       const startTime = Date.now();
       const maxDelay = configs.maximumAcceptableDelayForTabDuplication;
       while (Date.now() - startTime < maxDelay) {
-        newTabs = movedTabs.map(tab => Tab.get(TabIdFixer.fixTab(tab).id));
-        newTabs = newTabs.filter(tab => !!tab);
+        newTabs = movedTabs.filterMap(Tab.get(TabIdFixer.fixTab(tab).id) || undefined);
```
piroor: As reported at https://travis-ci.org/piroor/treestyletab/jobs/575602808 `tab => ` part looks missing.

webextensions/common/tabs-store.js
```diff
@@ -171,31 +171,38 @@ function* getMatchedTabsIterator(tabs, query, offset) {
...
-  for (const attribute of MATCHING_ATTRIBUTES) {
+  for (let i = 0, maxi = MATCHING_ATTRIBUTES.length; i < maxi; i++) {
+    const attribute = MATCHING_ATTRIBUTES[i];
+
     if (attribute in query &&
         !matched(tab[attribute], query[attribute]))
       return false;
-    if (`!${attribute}` in query &&
-        matched(tab[attribute], query[`!${attribute}`]))
+    const bangAttribute = `!${attribute}`
```
piroor: `inverted` instead of `bang` looks better in this context.

webextensions/common/tabs-store.js
```diff
@@ -171,31 +171,38 @@ function* getMatchedTabsIterator(tabs, query, offset) {
...
+    if (bangAttribute in query &&
+        matched(tab[attribute], query[bangAttribute]))
       return false;
   }
 
   if (!tab.$TST)
     return false;
 
-  if ('states' in query && tab.$TST.states) {
+  const tstStates = tab.$TST.states;
```
piroor: `tabStates` looks better than `tstStates` in this context.

webextensions/common/tabs-store.js
```diff
@@ -171,31 +171,38 @@ function* getMatchedTabsIterator(tabs, query, offset) {
 }
 
 function matchedWithQuery(tab, query) {
-  for (const attribute of MATCHING_ATTRIBUTES) {
+  for (let i = 0, maxi = MATCHING_ATTRIBUTES.length; i < maxi; i++) {
```
piroor: Very interesting. I didn't thought that "for-of" loop is worse than simple "for" loop. To prevent mis-reverting by me this in the future, could you add a comment (the commit message itself looks suitable) here?

webextensions/common/common.js
```diff
@@ -334,6 +337,41 @@ configs.$loaded.then(() => {
 });
 
 
+export function filterMap(arr, callback, thisArg) {
```
piroor: Could you use `array` or `items` instead of a shortened name `arr`, to uniform the coding style?

webextensions/common/common.js
```diff
@@ -334,6 +337,41 @@ configs.$loaded.then(() => {
...
+  arr = Object(arr);
+  const maxi = arr.length >>> 0;
+
+  if (typeof callback !== 'function')
+    throw new TypeError(`${callback} is not a function`);
+
+  callback = callback.bind(thisArg == undefined ? arr : thisArg);
+
+  const newArr = new Array(maxi);
+  let counti = 0,
```
piroor: `count` looks enough in this context. (The suffix `i` looks unnecessary.)

webextensions/common/common.js
```diff
@@ -334,6 +337,41 @@ configs.$loaded.then(() => {
...
+      i = 0,
+      value,
+      newValue;
+
+  while (i < maxi) {
+    if (i in arr) {
+      value = arr[i];
+      newValue = callback(value, i, arr);
+      if (newValue !== undefined) {
```
piroor: I'm afraid that I may forget to return `undefined` by the callback function, because I ordinarily ignore difference of `null`, `false` and `undefined`. Are `!newValue` or `newValue !== undegined && newValue !== false && newValue !== null` painfully expensive than `!== undefined`?

piroor:
Thanks! Very interesting. I excessively expected that the JIT compiler of Firefox itself may do such optimizations automatically. I think this should be merged even if these optimization is incomplete to reduce all bottlenecks.

I have some questions, so could you check my comments?

piroor:
I've realized that your `filterMap()` can be done with a builtin method of Array: `Array.prototype.reduce()`, as: https://github.com/piroor/treestyletab/commit/1d649f4dff917b17ff72889c39c7d3f8fe94ce4e
So I'm  trying to apply these changes with `reduce()`.

piroor:
Most parts of this PR was applied in different way, except conversion of a for-of loop to a for loop at: https://github.com/piroor/treestyletab/pull/2368/commits/b89d7519f25ee09af80ebb730b32a3d9942534e3

piroor:
On my environment "for-of" loop is 2-3 times faster than regular "for" loop. Benchmark:

```javascript
{
  const result = [];
  {
    let start = Date.now();
    let count = 0;
    for (let i = 0; i < 1000000; i++) {
      for (let j = 0, maxj = MATCHING_ATTRIBUTES.length; j < maxj; j++) {
        const attribute = MATCHING_ATTRIBUTES[j];
        count++;
      }
    }
    result.push(`for: ${Date.now() - start}`);
  }
  {
    let start = Date.now();
    let count = 0;
    for (let i = 0; i < 1000000; i++) {
      for (const attribute of MATCHING_ATTRIBUTES) {
        count++;
      }
    }
    result.push(`for-of: ${Date.now() - start}`);
  }
  console.log(result); // => Array [ "for: 2294", "for-of: 882" ]
}
```

I didn't compare RAM usages of these methods yet, but if "for-of" loop eats more RAM, which is the method I should choose?

piroor:
Sorry, I've finally understood that the importance of the logic of your `filterMap()` - it looks designed to reduce needless memory allocation more. My `Array.prototype.reduce()` version may trigger GC. I've introduced a more simplified version `mapAndFilter` (and other helpers) with the commit 4438f0c78.

ariasuni:
I’m also using a lot `filter` and `map` in Tab Center Reborn, so it could improve its performance.

Did you see a difference in practice with `while` instead of `for` and `undefined` instead of `null`?

I think `mapAndFilterUniq` lacks a comment to explain what it does because its code is bit difficult to understand.

Also, the three `// define as unsingned int` have a typo: it’s unsigned.

ariasuni:
I made a few (6 or so) different tests to see what’s more efficient. It showed me that simpler is often better.

Some context: tabs are stored as a `Map` in Tab Center Reborn (with id as a key, because we often want to get a tab by id), so I wanted to use an iterator over my `Map`’s values instead of converting it to an Array first, and such avoiding the creation of a temporary array.

This is what I coded:

```js
// return an array of values filtered and mapped (combine filter and map)
// mapper should have the form: filterCondition ? mappedValue : undefined
export function filterMap(iter, mapper, max) {
  const mappedValues = new Array(max);
  let count = 0;
  for (const value of iter) {
    const mappedValue = mapper(value);
    if (mappedValue !== undefined) {
      mappedValues[count] = mappedValue;
      count++;
    }
  }
  mappedValues.length = count;
  return mappedValues;
}

// return the first matching result, can be used to see if any value is matching
// mapper should have the form: filterCondition ? mappedValue : undefined
export function filterMapFind(iter, mapper) {
  let mappedValue;
  for (const value of iter) {
    mappedValue = mapper(value);
    if (mappedValue !== undefined) {
      return mappedValue;
    }
  }
  return null;
}

// return the count of matching values
// matcher should return a boolean
export function filterMapCount(iter, matcher) {
  let count = 0;
  for (const value of iter) {
    if (matcher(value)) {
      count++;
    }
  }
  return count;
}
```

piroor:
I've done more simplification at d8ebdf17e24b8dc4247b55ab691ecb3e2a3c7e13. `while` loops are replaced with `for`-`of` loops, but the loop variables are defined outside of each loop - this is the only one difference from https://github.com/piroor/treestyletab/pull/2368#issuecomment-531988927

piroor:
Anyway we need a simple and reproducible benchmark to compare effects of this "optimization"...

bb010g:
This is pretty stale.

PR #2381, opened 2019-09-13
Update Russian translation v3

wvxwxvw:
And if so


piroor:
Thank you for patient retryings!

Issue #2000, opened 2018-08-19
Should we use WebComponents for building a DOM element?


tetsuharuohzeki:
[From Firefox 63](https://groups.google.com/d/topic/mozilla.dev.platform/RcanREo7hU0/discussion), Gecko eventually [(will) ships](https://hg.mozilla.org/mozilla-central/rev/16daef3a89a2) Shadow DOM and Custom Element as known as [Web Components](https://developer.mozilla.org/en-US/docs/Web/Web_Components).

These features provides a way to create a custom DOM element and encapsulate it.
I think they would useful to building a view for this add-on.

irvinm:
@piroor can you comment on this? (Sorry, I don't have a sense if this is still applicable today or not)

piroor:
This was impossible on TST 2.x, but currently TST 3.x is designed very similar to this, so now it may be migratable to WebComponents. Especially if the migration improve performance or maintainability, I agree to do it.

saneyuki:
To support TST's _theme_ completly, I suspect it's better to wait [CSS Shadow Parts](https://drafts.csswg.org/css-shadow-parts-1/) ([bug 1551113](https://bugzilla.mozilla.org/show_bug.cgi?id=1551113)).

But I think we can replace a simple element with custom element from now.

saneyuki:
[CSS Shadow Parts](https://drafts.csswg.org/css-shadow-parts/) will [ships](https://groups.google.com/d/topic/mozilla.dev.platform/dWnXJtFKY0I/discussion) in Firefox v72.

By CSS Shadow Parts, I think we can provide a more reasonable way for style customization instead of depending on CSS variables. I suspect we can provide customizable style encapsulation.

irvinm:
With FF72 out, is there more to look at now?

piroor:
This looks should be progressed experimentally on a branch, until Firefox ESR68 becomes outdated. Maintaining multiple branches is hard for me, so I hope to concentrate to maintain and hope @saneyuki gives me experimental implementation examples...

saneyuki:
I'm sorry. I don't have a concrete progress on https://github.com/piroor/treestyletab/pull/2427.

Because:

1. I'm busy recently....
2. I suspect we might need a bunch of refactoring to implement [the proposed plan](https://github.com/piroor/treestyletab/pull/2427#issuecomment-566461006).
3. By _reason 2_, I felt it might be better we should wait the end of life for ESR68 (Mozilla releases new ESR).

[...]

PR #2392, opened 2019-10-15
Use a custom element to implement tab closebox 

saneyuki:
* This is a trial implementation for https://github.com/piroor/treestyletab/issues/2000.
* In this PR, I tried to replace most minimal piece so that it would be easiest work, proof of concept, and as the first step.
    * In this PR, we don't use shadow DOM. This only changes a local element to custom element.


webextensions/sidebar/components/TSTCloseBoxElement.js
```diff
@@ -0,0 +1,47 @@
...
+    this._isInitialized = false;
+  }
+
+  connectedCallback() {
+    if (this._isInitialized) {
+      return;
+    }
+
+    this.setAttribute('title', browser.i18n.getMessage(NORMAL_TOOLTIP));
+    this.setAttribute('draggable', true); // this is required to cancel click by dragging
```
saneyuki: * If we do these operations in a `constructor` of this class, Gecko throws `NotSupportedError: Operation is not supported`. 
* `connectedCallback()` may be called multiple times [by append/remove operations](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements).
* `browser.i18n.getMessage()`might be a costly operation.

So I make ensure to call these operation only once conservatively.

saneyuki:
@piroor how about do you think?

webextensions/common/constants.js
```diff
@@ -319,3 +318,5 @@ export const kSHORTHAND_URIS = {
 };
 
 export const kNOTIFICATION_DEFAULT_ICON = '/resources/icon64.png';
+
+export const kTAB_CLOSE_BOX_ELEMENT = 'tst-tab-closebox';
```
piroor: The prefix `tst-` looks unnecessary for me, because the namespace and the document namespace is already isolated for TST.

saneyuki: @piroor okay. I removed it in https://github.com/piroor/treestyletab/pull/2392/commits/16beebe03857d0162b83e8a42f0ab04f232f73f9

webextensions/sidebar/components/TSTCloseBoxElement.js
piroor: Thank you for explanation! This information (especially the first one) looks good to be written as a comment with the URL of the corresponding bug on the BMO.

webextensions/sidebar/components/TSTCloseBoxElement.js
```diff
@@ -0,0 +1,47 @@
+const NORMAL_TOOLTIP = 'tab_closebox_tab_tooltip';
+const MULTISELECTED_TOOLTIP = 'tab_closebox_tab_tooltip_multiselected';
+const TREE_TOOLTIP = 'tab_closebox_tree_tooltip';
+
+export const CloseBoxTooltipType = Object.freeze({
+  Normal: 'normal',
+  MultiSelected: 'multiselected',
+  Tree: 'tree',
```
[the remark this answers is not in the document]
saneyuki: > Is the PascalCase common for a usecase like this? (Sorry I'm not familiar to common coding styles in WebComponents.)

Ah, I wrote them to reduce an exported object and imagined the style of TypeScript's _enum_ feature.
Should I change to UPPER_CASE?

webextensions/sidebar/components/TSTCloseBoxElement.js
saneyuki: I'm not familiar with details of the spec, but this is not Gecko's bug.

> 6. If result has children, then throw a "NotSupportedError" DOMException. 
>
> https://dom.spec.whatwg.org/#concept-create-element

webextensions/sidebar/sidebar-tabs.js
```diff
@@ -41,6 +43,13 @@ export const wholeContainer = document.querySelector('#all-tabs');
 export const onSyncFailed = new EventListenerManager();
 
 export function init() {
+  const elements = [
+    [Constants.kTAB_CLOSE_BOX_ELEMENT, TSTCloseBoxElement],
+  ];
+  for (const [localName, constructor] of elements) {
+    window.customElements.define(localName, constructor);
+  }
+
```
piroor: Why you didn't put these codes in the `TSTCloseBoxElement.js`? It looks better for me to define the relation of `Constants.kTAB_CLOSE_BOX_ELEMENT` and `TSTCloseBoxElement` in the `TSTCloseBoxElement.js` itself. If there is any reason why we shouldn't do that, I need to understand it well. My understanding is:

* The element name may be changed in cases, so we should not tie them as a pair in general.
* When I reuse these codes partially for other addons, tightly paired element name may block us to do that.

If they are full reasons of this design, then I still think that the element constructor class can have its default element name as something like its public property or others.

webextensions/common/constants.js
piroor: Oops, with the same reason `TST` prefix for classes and file names also kook unnecessary for me. (Did you give the prefix for a future usecase reusing the file in other addons?)

webextensions/sidebar/components/TSTCloseBoxElement.js
piroor: Thanks, I misunderstood that the "NotSupported" means "Gecko does not support it". I got that it actually means "such an operation is not supported in the spec." Then I think that a comment with the URL of the spec may help future work for me until I master the WebComponents spec completely.

webextensions/common/constants.js
saneyuki: > Oops, with the same reason TST prefix for classes and file names also look unnecessary for me. (Did you give the prefix for a future usecase reusing the file in other addons?)

Sorry. I forgot to rename them. I'll change in the next patch.

webextensions/sidebar/components/TSTCloseBoxElement.js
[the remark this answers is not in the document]
piroor: OK, I've understood the context. I'm OK to use PascalCase for enum-like codes in this project, for a future migration to TypeScript (I don't know I can do it in near future...) And (this is off topic) with the same reason other enum-like constants in this project should be re-defined in this style for consistency - it will be separated to another PR.

webextensions/sidebar/sidebar-tabs.js
saneyuki: > Why you didn't put these codes in the TSTCloseBoxElement.js?

If we call `window.customElements.define(localName, constructor);` from the file defining a custom element, it would be a side effect and happen accidentally that defining a custom element when we import a new file which defines a new custom element. It causes a complex side effect relations. 

I have not concluded the best practice about it yet, but I think that it's safely to call window.customElements.define(localName, constructor);` separately in the application initialization phase.

> The element name may be changed in cases, so we should not tie them as a pair in general.

I did not intend to make a local name and a class loose coupled.

I'll add the change to move `Constants.kTAB_CLOSE_BOX_ELEMENT` to the file which defines the custom element.

webextensions/sidebar/components/TSTCloseBoxElement.js
saneyuki: I've added the comment: https://github.com/piroor/treestyletab/pull/2392/commits/3402277a619631f7f258afa430a0d0ad084235c0

webextensions/common/constants.js
saneyuki: I've renamed them https://github.com/piroor/treestyletab/pull/2392/commits/fae20fbb29ef1f7a62da47550547f4b22b9f64cc

webextensions/sidebar/sidebar-tabs.js
piroor: OK, I agree that accidental side effect is not good and `window.customElements.define()` should be here. 

> I'll add the change to move Constants.kTAB_CLOSE_BOX_ELEMENT to the file which defines the custom element.

Thanks. Defining the constant for the element name in the same file looks reasonable. I'm imagine that it is accessible like `TabCloseBoxElement.ELEMENT_NAME` or something, but it is just a matter of personal preference. (But I can't understand why the spec doesn't provide a way to resolve the default element name from the class name itself automatically...)

saneyuki: @piroor 

I updated the patch https://github.com/piroor/treestyletab/pull/2392/commits/21f4fa9c1fa09784cf8c1b97ee76f2eb846fbdb1

> I think that it's rare case that we need static properties/methods for class in today's ECMAScript.

I believe we fulfill this "rare case" by both of them

1. We can provide an ergonomic API.
2. A provided operation/values has a strong relation to its data class.

piroor:
Hmm...? GitHub doesn't detect "merged" status automatically. I don't know the reason, but I close this anyway because this has been really merged.

piroor:
Anyway this will be a good guideline to do this type migration by myself. Thanks!

saneyuki:
You're welcome!

PR #2393, opened 2019-10-16
Use a custom element to implement tab-favicon

saneyuki:
* This simply replace `span` element to `tab-favicon` custom element.
* Ideally, these descendants should be in shadow tree. Thus I don't change these element to custom elements. But I hesitate to do it at this moment by these reasons.
    * If we move these to shadow tree,
        * We need some rewrite our style.
            * This includes that we need to move almost CSS code into this file as a string.
            * I'm not sure about the decision to move CSS code as a string into each of custom element file. Of course, there are some pros/cons. 
        * I'm not sure about that whether we should require [CSS Shadow Parts](https://bugzilla.mozilla.org/show_bug.cgi?id=1559074).
        * I suspect we can resolve almost problems by using CSS Custom Properties.

## Related Issue

https://github.com/piroor/treestyletab/issues/2000

### DIFF
--- a/webextensions/sidebar/styles/favicon.css
+++ b/webextensions/sidebar/styles/favicon.css
@@ -5,7 +5,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 */
 
-.favicon {
+tab-favicon {
   display: inline-block;
   font-size: var(--favicon-size);
   height: var(--favicon-size);
@@ -19,11 +19,11 @@
   white-space: pre;
   width: var(--favicon-size);
 }
-.tab .favicon,
-.tab .favicon * {
+.tab tab-favicon,
+.tab tab-favicon * {
   vertical-align: baseline;
 }
-.tab .favicon * {
+.tab tab-favicon * {
   position: relative;
   z-index: 20;
 }
--- a/webextensions/sidebar/styles/favicon.css
+++ b/webextensions/sidebar/styles/favicon.css
@@ -107,7 +107,7 @@ tab-favicon {
 }
 
 
-.tab.group-tab .favicon-image[src]:not(.error) {
+.tab.group-tab tab-favicon:not(.error) .favicon-image[src] {
   bottom: 0;
   max-height: 10px;
   max-width: 10px;
@@ -117,14 +117,14 @@ tab-favicon {
 }
 
 
-.favicon-image.error,
+tab-favicon.error .favicon-image,
 .favicon-image:not([src]),
 .tab[data-current-uri^="chrome:"] .favicon-image,
 .tab[data-current-uri^="about:addons"] .favicon-image,
 .tab[data-current-uri^="about:debugging"] .favicon-image,
 .tab[data-current-uri^="about:preferences"] .favicon-image,
 .tab:not(.group-tab):not([data-current-uri^="chrome:"]):not([data-current-uri^="about:addons"]):not([data-current-uri^="about:debugging"]):not([data-current-uri^="about:preferences"])
-  .favicon-image[src]:not(.error) ~ .favicon-builtin::before,
+  tab-favicon:not(.error) .favicon-image[src] ~ .favicon-builtin::before,
 .tab.loading .favicon-image,
 .tab.loading .favicon-builtin::before {
   display: none !important;
--- a/webextensions/common/constants.js
+++ b/webextensions/common/constants.js
@@ -6,7 +6,7 @@
 'use strict';
 
 export const kBACKGROUND_CONTENTS_VERSION = 5;
-export const kSIDEBAR_CONTENTS_VERSION    = 10;
+export const kSIDEBAR_CONTENTS_VERSION    = 11;
 
 export const kCOMMAND_GET_INSTANCE_ID                = 'treestyletab:get-instance-id';
 export const kCOMMAND_RELOAD                         = 'treestyletab:reload';
--- a/webextensions/common/constants.js
+++ b/webextensions/common/constants.js
@@ -113,7 +113,6 @@ export const kPERSISTENT_SUBTREE_COLLAPSED = 'subtree-collapsed'; // obsolete
 export const kPERSISTENT_ORIGINAL_OPENER_TAB_ID            = 'data-original-opener-tab-id';
 export const kPERSISTENT_ALREADY_GROUPED_FOR_PINNED_OPENER = 'data-already-grouped-for-pinned-opener';
 
-export const kFAVICON         = 'favicon';
 export const kFAVICON_IMAGE   = 'favicon-image';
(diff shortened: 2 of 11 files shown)


saneyuki:
@piroor 

How about this?
By this and https://github.com/piroor/treestyletab/pull/2392, I feel there are some lessons to migrate to a custom element from simple DOM.

1. It's easy to replace `span`/`div` with a custom element and remove needless `class` attribute.
    * This is simple and we don't have to care about shadow tree boundary.
    * But there are some elements which we should encapsulate in Shadow DOM.
1. Be careful to _move a view logic which updates itself to a custom element_.
    * This relates fundamental _view_ architecture design of this application.
    * If we do this, we might tweak about the order of _view_ initialization and defining custom elements.
        * It could be a bit large refactoring.

piroor:
I've understood that shadow DOM cannot be introduced until the CSS Shadow Parts is activated by default.

> Be careful to move a view logic which updates itself to a custom element. 

Should codes to update favicons be moved into the definition of the `TabFaviconElement` class in near  future? I'm sorry but I'm not well about recent trend of Web technologies and I understand just traditional MVC architecture introduced in Ruby on Rails. On such a traditional MVC, the present design of TST looks like a fat controller (and of course it is a bad practice.) Even if this PR is merged the fat controller will still there. Do you mean that the fat controller should be split and codes to update faivcons should be cupled in the `TabFaviconElement` class as the "next" step after major custom CSS classes are replaced with custom elements? If this understanding is correct, it will require more hard work but we must do that for more stability and maintainability, and I agree that it is a required task.

----

Hmm, by the way you looks to need to update more elements for example: https://github.com/piroor/treestyletab/blob/a084d1cb6946f28423827ddfdb83b78156b3a7fc/webextensions/sidebar/sidebar.html#L98
The dummy element is required to calculate actual size of tab related elements consistently. And as you saw the dummy element for favicons contains the master throbber to synchronize animations of other throbbers. Thus I think you need to rewrite related codes carefully... (I think the master throbber should be re-defined separately from the dummy favicon.)

saneyuki:
> If we do this, we might tweak about the order of view initialization and defining custom elements. 

For example, I changed https://github.com/piroor/treestyletab/pull/2393/commits/8ea3776990c9b242c8d5c7261ef4765f8c4d3011 to implement [TabFaviconElement.getImageElement()](https://github.com/piroor/treestyletab/pull/2393/commits/73ef07df397e1d6be8413b9673db97858f40a4db).
I think this would resolve almost problems related to initialization order between a custom element and other logic. But it's a far from the place calling `SidebarTabs.init()`.

saneyuki:
> Do you mean that the fat controller should be split and codes to update faivcons should be cupled in the TabFaviconElement class as the "next" step after major custom CSS classes are replaced with custom elements? If this understanding is correct, it will require more hard work but we must do that for more stability and maintainability, and I agree that it is a required task.

Ah, sorry, I don't have concrete opinion yet.

Ideally, I think it's better to break down the current fat controller design. But I also think it would be a tough work definitely.

According to the recent practices of (web) GUI application programming, we should make _view_ layer thin. If we move codes to update faivcons to `TabFaviconElement`, we need care about that `TabFaviconElement` should not be _smart_ (e.g. IMO, `TabFaviconElement` does not know how to resolve a favicon url). It also would be a tough work.

saneyuki:
> Hmm, by the way you looks to need to update more elements for example: 

I addresses in https://github.com/piroor/treestyletab/pull/2393/commits/f5defaf100a3e54919ddbe228b055dbb976944f6 & https://github.com/piroor/treestyletab/pull/2393/commits/609098c2b354572b5ce45db2409bba7365072454 (I squashed)

piroor:
Thank you for explanation. You're right, such tough work will increment the major version of TST to 4.0...

Anyway this type capsulization allows us to remove `setAttribute()` and other initialization codes from the main logic, so I have no objection to migration to custom elements. Just this migration should simply increment the minor version to 3.2.

Two more things could you apply?

1. `kSIDEBAR_CONTENTS_VERSION` need to be incremented when DOM structure of the sidebar contents is changed. https://github.com/piroor/treestyletab/blob/a084d1cb6946f28423827ddfdb83b78156b3a7fc/webextensions/common/constants.js#L9 Without this change, TST wrongly restores old cached contents from a saved HTML source so restored tabs are shown with broken appearance. (By the way this constant looks should be moved to sidebar/sidebar-cache.js because no other module refers it.)
2. Even if it is unnecessary in TST itself, old class names of span/div should be added to custom elements. I agree `<tab-closebox class="closebox">` looks odd but this improves backward compatibility with other addons (like Multiple Tab Handler) and user stylesheets.

saneyuki:
> kSIDEBAR_CONTENTS_VERSION need to be incremented when DOM structure of the sidebar contents is changed. 

I addressed it in https://github.com/piroor/treestyletab/pull/2393/commits/7040793d915e3039918ba2ef9b0dca8a226b9618 (fixup & squashed)

> Even if it is unnecessary in TST itself, old class names of span/div should be added to custom elements. I agree <tab-closebox class="closebox"> looks odd but this improves backward compatibility with other addons (like Multiple Tab Handler) and user stylesheets.

I added the commit to preserve their classes in https://github.com/piroor/treestyletab/pull/2393/commits/02e79f86c0205d7244e76fcb58f06610cbd8b92b

piroor:
Thank you for updating.

After trying I've seen a regression: favicon of tabs are not restored after restarting. Currently last favicon images are saved like as `` in the cached HTML source. With this change the `src` attribute of the img element looks to be cleared always. Thus I think information currently stored at descendant elements need to be stored the tab-favicon element itself and applied to its descendants, like XBL did.

saneyuki:
> Thus I think information currently stored at descendant elements need to be stored at the tab-favicon element itself and applied to its descendants, like XBL did.

I tried to addess it in https://github.com/piroor/treestyletab/pull/2393/commits/be71d04eeda14ff0cd7022ffebf12487a7ade17b
How about this?

webextensions/sidebar/styles/favicon.css
```diff
@@ -19,11 +19,11 @@
   white-space: pre;
   width: var(--favicon-size);
 }
-.tab .favicon,
-.tab .favicon * {
+.tab tab-favicon,
+.tab tab-favicon* {
```
piroor: A space is missing between `tab-favicon` and `*`.

webextensions/sidebar/styles/favicon.css
```diff
@@ -19,11 +19,11 @@
...
   vertical-align: baseline;
 }
-.tab .favicon * {
+.tab tab-favicon* {
```
piroor: A space is missing between `tab-favicon` and `*`.

saneyuki:
@piroor I addressed them

piroor:
Thanks!

PR #2398, opened 2019-10-26
Avoid unnecessary shape transition for custom elements

saneyuki:
`delete` causes to change a Shape (SM)/Hidden Class (V8)/Structure (JSC) in JSVM. They treats an object which we remove a property from by `detele` as _dictionary mode_ internally.

If we treat the object as dictionary or hashmap by adding/removeing properties frequently, this JSVM behavior is nice. But it would not be effectively if we treats the object as normal class.

Practically, to avoid unnecessary this kind of internal behavior and to stabilize object's shape,

1. We should initialize all properties with the same order for each of times.
    * _constructor_ function is nice place.
2. We should fill with empty value (`null`, `undefined`, empty string `''`, `-1`, or others) when we would not like to hold some values in a property.

### DIFF
--- a/webextensions/sidebar/components/TabCloseBoxElement.js
+++ b/webextensions/sidebar/components/TabCloseBoxElement.js
@@ -19,6 +19,7 @@ export class TabCloseBoxElement extends HTMLElement {
 
   constructor() {
     super();
+    this._reservedUpdate = null;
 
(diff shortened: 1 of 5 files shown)


piroor:
Thanks, I missed the point.

PR #2399, opened 2019-10-30
Update custom style rule example in options

andytham:
Nothing important, but updated the commented example in advanced options to reflect the current way to change unread tab styles.


piroor:
Thanks!

Issue #1363, opened 2017-09-22
Change the styling of unloaded (pending) tabs


Keith94:
Is it possible to change the css for unloaded tabs yet? For example change opacity, etc..

piroor:
By recent commits, now discarded tabs have "discarded" class. So user-defined style rule like following will work as expected:

```css
.tab.discarded { opacity: 0.75; }
```

piroor:
This is also filed at the code snippets:
https://github.com/piroor/treestyletab/wiki/Code-snippets-for-custom-style-rules

kkot:
@piroor Could you tell me how to make only icon more opaque ?

vn971:
Great to see it working!
I think it's important enough to be built-in.

Ainoss:
Unfortunately, it doesn't work for me after upgrading to 57.0.2 (or this is just coincidence). 
User-defined style rule as @piroor posted above does nothing.
Is it a known problem, or just for me? It will be great if anybody could give solution.

vn971:
@Ainoss overriding font somewhy stopped working for me (I override with   `color: red;`), but opacity works. I'm on ArchLinux, FF version 57.0.2.

encukou:
works with:
```
.tab.discarded .label {
  color: red;
}
```

Ainoss:
@encukou Thank for your suggestion! It works fine (win10).

alreadydone:
Unable to make this work (on Windows 10, FF60.0b12) with any of the above approaches ... any ideas? Not really familiar with CSS, but "Hide the 'Tree Style Tab' header" and "Hide horizontal tabs at the top of the window" from https://github.com/piroor/treestyletab/wiki/Code-snippets-for-custom-style-rules#for-userchromecss were successful.

piroor:
@alreadydone You need to put custom CSS rules to an input field in TST's configuration page, instead of your userChrome.css file.

alreadydone:
It works. Thanks!

gnomesley:
This seems to be broken in the latest version (2.4.19) with FF 60.

piroor:
@gnomesley Please add one more extra CSS rule:

```css

.tab .label-content {
  color: inherit;
}
```

I'll include this as a builtin rule at the next release of TST. Sorry.

gnomesley:
Thanks for the response!

bb010g:
`userChrome.css` equivalent to the wiki TST snippet:

```css
[...]

PR #2427, opened 2019-12-04
Expose span element in tab-label element with CSS Shadow Parts

tetsuharuohzeki:
## Motivation

* This is an experiment to use [CSS Shadow Parts](https://drafts.csswg.org/css-shadow-parts/), which Firefox will ship in v72.
* By CSS Shadow Parts, we can expose some elements in a shadow tree with strong encapsulation.
    * By this, we can provide a public API to customize style.

## Drawback or Considerations

1. User would not apply their custom style sheet without any limitation.
    * IMO, I doubt user requires that all elements should be exposed. But I'm not sure about TST's custom style sheet. My thought may be wrong.
2. This requires to bump up `strict_min_version` to `72a1`.
    * This means we cannot deliver our bugfix to Firefox ~71 which Mozilla plans to release in [2020-01-09](https://wiki.mozilla.org/Release_Management/Calendar).

## Tasks after merge this PR

1. We may need to update https://github.com/piroor/treestyletab/issues/1363 or other documents to note about this change.

## Related Issue

* https://github.com/piroor/treestyletab/issues/2000
* https://github.com/piroor/treestyletab/issues/1363

### DIFF
--- a/webextensions/sidebar/components/TabLabelElement.js
+++ b/webextensions/sidebar/components/TabLabelElement.js
@@ -7,7 +7,6 @@
 export const kTAB_LABEL_ELEMENT_NAME = 'tab-label';
 
 const KLABEL_CLASS_NAME   = 'label';
-const kCONTENT_CLASS_NAME = `${KLABEL_CLASS_NAME}-content`;
 
 const kPART_LABEL_TEXT = 'label-text';
 
@@ -35,7 +34,6 @@ export class TabLabelElement extends HTMLElement {
 
     const label = document.createElement('span');
     shadow.appendChild(label);
-    label.classList.add(kCONTENT_CLASS_NAME);
     label.setAttribute('part', kPART_LABEL_TEXT);
   }
 
@@ -107,7 +105,7 @@ export class TabLabelElement extends HTMLElement {
   }
 
   get _content() {
-    return this.shadowRoot.querySelector(`.${kCONTENT_CLASS_NAME}`);
+    return this.shadowRoot.querySelector(`[part=${kPART_LABEL_TEXT}]`);
   }
 
   get value() {
--- a/webextensions/sidebar/components/TabLabelElement.js
+++ b/webextensions/sidebar/components/TabLabelElement.js
@@ -9,6 +9,8 @@ export const kTAB_LABEL_ELEMENT_NAME = 'tab-label';
 const KLABEL_CLASS_NAME   = 'label';
 const kCONTENT_CLASS_NAME = `${KLABEL_CLASS_NAME}-content`;
 
+const kPART_LABEL_TEXT = 'label-text';
+
 const kATTR_NAME_VALUE = 'value';
 
 export class TabLabelElement extends HTMLElement {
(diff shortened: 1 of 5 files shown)


webextensions/sidebar/styles/base.css
```diff
@@ -385,7 +385,7 @@ tab-item tab-label {
 }
 
 /* This is for backward compatibility about custom user style rules like https://github.com/piroor/treestyletab/issues/1363 */
-tab-item .label-content {
+tab-label::part(textspan) {
```
saneyuki: _textspan_ is nice name?

saneyuki:
@piroor How do you feel about this?

webextensions/sidebar/components/TabLabelElement.js
```diff
@@ -99,7 +105,9 @@ export class TabLabelElement extends HTMLElement {
   }
 
   get _content() {
-    return this.querySelector(`.${kCONTENT_CLASS_NAME}`);
+    const shadow = this.shadowRoot;
```
piroor: <code>return this.shadowRoot.querySelector(`[part=${kPART_TEXT_SPAN}]`)</code> looks enough short here.

webextensions/sidebar/styles/base.css
piroor: The term "span" looks just to mean "a range". I think "label-text" looks more meaningful here.

piroor:
Thanks you for an experimental implementation! I've tried a user stylesheet:

```css
tab-item,
tab-item tab-label {
  color: red;
  font-style: italic;
  font-size: large;
}
```

and I've realized that it works completely as I expected. I hope that this kind changes may make TST's tabs more easily configurable.

Basically I think that TST should support the latest Firefox ESR. The next ESR is 78 [scheculed to be released at 2020-06-30](https://wiki.mozilla.org/Release_Management/Calendar), so if I drop support for ESR68, ESR users may need to stay on an old version of TST for half a year. I think this kind restructurings will improve maintainability but it looks not enough to be a reason to drop ESR support. So I think this kind work should be done just on a branch until ESR78 is out, but I'm also afraid that it become unmergable due to large changes on the master branch. How do you think about the migration and merge plan?

saneyuki:
> So I think this kind work should be done just on a branch until ESR78 is out, but I'm also afraid that it become unmergable due to large changes on the master branch. How do you think about the migration and merge plan?

Umm, it's difficult problem...

IMO, I don't think it would be a problem to bump up our minimum supported version to it after ESR 78 has been released because user would update their Firefox to 78 in the half year. 
But it's a problem to wait ESR 78 until 2020-06-30.

I'll close this PR once.

--------------

For internal refactoring purpose, I have an another approach to clean up exist CSS slector with BEM or similar style [like this approach](https://speakerdeck.com/kubosho/solution-of-css-is-too-fragile-by-abematv).

In such approach, followings are basic principles.

## Basic principles

* We use redundant _unique_ selector (e.g. `.sidebar-TabFaviconElement--img__state_barfoo`).
* We keep current class attributes for backward compatibility.
* Create a css file for each of components (custom elements).
    * Load css file by `@import url(...)` instead of css pre-processor.

## Pros

* We can refactoring exist codes.
   * We don't touch new CSS features which we cannot treats as progressive enhancement.
* For the future, when we start to use Shadow DOM, we may migrate the code to smooth transition.
    * By this change, we can know where override a style of descendant components.

## Cons

* User might write their custom style sheets by new selector rules.
    * It breaks user custom style sheets when we refactor our code for the future.
       * This problem would be happen if we move to CSS shadow parts.
* When we stop to keep backward compatibility for user custom style sheets?

PR #2434, opened 2019-12-21
Update Chinese Localization

no1xsyzy:
complete newly added strings
add spaces around quote marks `_“` or `”_` unless adjacent to another punctuation, for better typesetting with Microsoft YaHei.
make the chinese name `树状标签页`, which I would prefer, which can be discussed more.

### DIFF
--- a/webextensions/_locales/zh_CN/messages.json
+++ b/webextensions/_locales/zh_CN/messages.json
@@ -1,18 +1,18 @@
 {
     "extensionName": {
-        "message": "树状标签页",
+        "message": "Tree Style Tab - 树状标签页管理",
         "hash": "94c80eccdd67c127e1de28b71af57aee"
     },
     "extensionDescription": {
-        "message": "树状显示标签页。",
+        "message": "在侧栏中以树状图形显示标签页。",
         "hash": "1cf26efef213812fa419388efe44fea2"
     },
     "sidebarTitle": {
-        "message": "树状标签页",
+        "message": "Tree Style Tab",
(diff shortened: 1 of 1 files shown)


piroor:
Thanks!

>  make the chinese name 树状标签页, which I would prefer, which can be discussed more.

Hmm... it's a complecated topic. For existing users `Tree Style Tab (树状标签页)` looks to have better findability, because they already seen TST with the English name. On the other hand, localized name looks better for new users. (To be honest I'm in regret about this point for the `ja` locale, because I just put localized name in it.)

no1xsyzy:
@piroor after created the PR, I checked AMO and found that there are too few addons with Chinese characters. Most of these also keep their brand-ish part of name and append a localized one-phrase description. So I would go that way and set the name as `Tree Style Tab - 树状标签页管理`.

piroor:
Oh, sorry I meant about only the `extensionName` field. I agree that localized `sidebarTitle` looks good. So after I merge this I'll re-apply some of changes you reverted. Anyway thanks a lot!

piroor:
I've done changes to apply a new policy about localization to zh_CN and ja locales.

* Unlocalized: the name of this addon itself
* Localized: the title of the sidebar panel and the toolbar button

commits:

* fca93a20030a6390f0065c7eeb44e0d504da66f6
* 73cee9a81882b1f09de149e8a549620b24fe31e7

How do you think about this? I think this looks natural when I see a list of switchable sidebar panels (because all other panels provided by Firefox itself are also localized), but I'm afraid that some people get confused from the mismatch  between the name of this addon and the title of the sidebar panel.

PR #2446, opened 2020-01-03
Remove TabIdFixer

adroitwhiz:
It was a workaround for a [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1398272) fixed a year and a half ago. The Firefox fix should now be in all supported Firefox versions, so `TabIdFixer` is unnecessary.


piroor:
Thanks!

PR #2518, opened 2020-03-13
Update German translation

SammysHP:
These strings were missing in the last update because the English
strings were included in the German language file.


piroor:
Thanks!

PR #2539, opened 2020-03-31
Add/fix keyboard shortcut

no1xsyzy:
After right click a tab in sidebar, press T to expand to TST sub-menu.

The translation accidentally and dramatically broke my workflow `Right click - key T - key C` to close a tree. It seems Firefox is using first letter as a default shorthand, therefore, translation breaks that.


piroor:
Thanks!

PR #2541, opened 2020-04-01
Add TST-DisableMiddleClick to list of TST addons

adamsmd:
This adds [TST-DisableMiddleClick](https://addons.mozilla.org/en-US/firefox/addon/tst-disablemiddleclick/) to the list of TST addons in the README.md.

Disclaimer: I have no association with TST-DisableMiddleClick. It is just a TST addon that I use and find useful.


piroor:
Thanks, I didn't know it.

PR #2575, opened 2020-05-03
Update messages.json RUS

wvxwxvw:


piroor:
Thanks! But now I'm trying to extract the new feature to a separate addon, so this will be moved to it...

wvxwxvw:
Oh well. I'm just in a hurry, I guess. :)

PR #2679, opened 2020-08-05
Move "Color scheme" option from Development to Appearance section

adroitwhiz:
I had the same issue as everyone else did, where Firefox doesn't seem to provide the correct colors when using dark themes on Linux.

It turns out that using the Photon color scheme fixes this and makes the sidebar appear dark again, but that option was hard to find because it was placed in the Development section with debugging options.

Since it affects the tab bar's appearance, I think it makes more sense to put it in the Appearance section. I'm not sure why it was put in Development--does it have no effect on non-Linux platforms or break them?

Before | After
-- | --
image| image

### DIFF
--- a/webextensions/options/options.html
+++ b/webextensions/options/options.html
@@ -83,6 +83,16 @@
               __MSG_config_style_none_info__</label></li>
       </ul>
     </fieldset>
+    <p>__MSG_config_colorScheme_caption__
+      <label class="has-radio inline"
+            ><input type="radio"
+                    name="colorScheme" value="photon">
+        __MSG_config_colorScheme_photon__</label>
+      <label class="has-radio inline"
+            ><input type="radio"
+                    name="colorScheme" value="system-color">
+        __MSG_config_colorScheme_systemColor__</label>
+   </p>
     <p><label class="has-checkbox"
              ><input id="animation"
                      type="checkbox">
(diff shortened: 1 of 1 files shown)


piroor:
I'm negative about exposing the option to all users, instead I think we should do efforts to improve compatibility of the theme to the combination Linux + dark theme.

I tested again on Ubuntu 20.04 + "Adwaita Dark" with the system color scheme and I've realized that system colors used by addons don't respect the actual color of the theme of the OS itself. Thus now I think we should apply the system colors only on the combination: Linux with non-dark theme.

piroor:
Could you try the latest development build https://github.com/piroor/treestyletab#development-builds including the commit e65cf22 ?

hotsphink:
Possibly related: https://bugzilla.mozilla.org/show_bug.cgi?id=1667832#c1 which points to a phabricator revision with the comment:

This is intentionally not done on Linux: https://bugzilla.mozilla.org/show_bug.cgi?id=1529897

The reason is that Linux dark mode detection is based on luminance of certain system colors, which can be ambiguous for some themes (e.g. themes that are in the middle of light/dark, themes which use dark colors for certain areas but light for others)

piroor:
Memorandum: on my Ubuntu 20.04LTS (VM on VirtualBox) + Firefox 86 + TST 3.7.2, TST's tabs are shown with the built-in color scheme same to Firefox's Dark theme for in-content UIs (ex. about:addons) as I expected, on the combination Platform=Dark(=Yaru-dark), Firefox=Default, and "system color scheme" at TST.

emilio:
Oh hah, does https://github.com/piroor/treestyletab/pull/3047 fix all this? :-)

piroor:
@emilio Thanks a lot! With the change I successfully got dark color sidebar without any extra customization on Firefox and TST (tested with Nightly 96.0a1). I think this can be closed.

PR #2733, opened 2020-10-03
Fix race condition when applying styles to sidebar

Lej77:
When testing my [Tree Style Tab in Separate Window](https://addons.mozilla.org/firefox/addon/tst-in-separate-window/) helper extension I noticed that the style is a bit messed up when the sidebar page is opened in a normal tab (and therefore also when it is opened in another window). To be precise I have dark theme enabled and the active tab is highlighted in white. I tracked the problem down to a race condition and confirmed that the change in this PR fixes it.

Sidebar page to the left and normal sidebar to the right:
Tst style issue

### DIFF
--- a/webextensions/sidebar/sidebar.js
+++ b/webextensions/sidebar/sidebar.js
@@ -97,6 +97,7 @@ const mContextualIdentitiesStyle  = document.querySelector('#contextual-identity
     configs.$loaded.then(async () => {
       await applyStyle();
       Size.update();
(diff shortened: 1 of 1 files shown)


piroor:
I missed that, thanks!

piroor:
Related to this topic, I've introduced more changes to append the "style" query parameter by TST side with the commit dca685c84. Now separate sidebar window will be shown with the chosen theme without any lag.

Lej77:
Wow, thanks! I really appreciate how well you support helper extensions for Tree Style Tab!

PR #2746, opened 2020-10-31
Rename korean translation directory from kr to ko

dfkdream:
Firefox uses `ko` for korean locale code, so I changed locale directory name to `ko`.


piroor:
Thanks!

PR #2748, opened 2020-11-04
Add missing contextual-identity fence icon

maciekmm:
The fence icon was missing from contextual-identities icons.

It is used for example by "Facebook Container" add-on.

There's still an issue when the color chosen is `toolbar`.


piroor:
Thanks! I'll try Facebook Container later.

piroor:
I've added more codes to support "toolbar" color on both dark and blight theme.

PR #2784, opened 2020-12-16
Register wheel event as `passive : false`

misonln41:
**Aim**
 - Make [Tree Style Tab Mouse Wheel](https://addons.mozilla.org/ja/firefox/addon/tree-style-tab-mouse-wheel/) works properly.
 - Tree Style Tab Mouse Wheel is an TST extend addon that enables users to switch active tab by scrolling mouse wheel on TST Sidebar, but now, this addon cannot handle scroll.
 - This TST extend addon calls Tree Style Tab API to handle scroll on sidebar, so this problem may not be fixed in extend addon side.

**Changes added**
 - Set `passive : false` in wheel event listener.
 - This change makes `event.preventDefault();` in `onWheel` event runs correctly.

**Risk**
 - This change may cause some degradation of rendering performance.
 - reference: https://developer.mozilla.org/ja/docs/Web/API/EventTarget/addEventListener#Improving_scrolling_performance_with_passive_listeners

**Environments**
 - Firefox 84.0
 - Latest version of Tree Style Tab / Tree Style Tab Mouse Wheel
 - KDE Plasma
 - Arch Linux

### DIFF
--- a/webextensions/sidebar/scroll.js
+++ b/webextensions/sidebar/scroll.js
@@ -71,7 +71,7 @@ export function init(scrollPosition) {
     });
   }
 
-  document.addEventListener('wheel', onWheel, { capture: true });
+  document.addEventListener('wheel', onWheel, { capture: true, passive: false });
   mTabBar.addEventListener('scroll', onScroll);
   BackgroundConnection.onMessage.addListener(onBackgroundMessage);
   TSTAPI.onMessageExternal.addListener(onMessageExternal);


misonln41:
This is the probrem happens in current TST and TST Mouse Wheel.

When I scroll mouse wheel, not only the tab changes, but also scrolls. Seems to be failing to handle scroll event.

Peek 2020-12-16 20-11

piroor:
Thanks a lot! The default "passive" option value looked to be changed to "true" on Firefox 84.

PR #2798, opened 2021-01-17
Update German translation

SammysHP:
I haven't understood the meaning of `config_groupTabTemporaryState_option_*`, even after having a look at the context. Maybe you can describe what it does so that I can find a proper translation.

### DIFF
--- a/webextensions/_locales/de/messages.json
+++ b/webextensions/_locales/de/messages.json
@@ -3,10 +3,8 @@
   "extensionDescription": { "message": "Stellt Tabs in einer Baumstruktur dar." },
 
   "sidebarTitle": { "message": "Tree Style Tab" },
-  "sidebarToggleDescription": { "message": "\"Tree Style Tab\" Sidebar Ein-/Ausblenden" },
+  "sidebarToggleDescription": { "message": "\"Tree Style Tab\" Seitenleiste Ein-/Ausblenden" },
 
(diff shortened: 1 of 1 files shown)


piroor:
Thank you for updating!

These messages are used for an grouped options under the "New Tabs Behavior" section. It is available only when you activate expert options:
image

It aims to control the checkbox state in "group tabs" opened by TST automatically. For example:
image
In this screenshot there are two pinned tabs, and I open two tabs from links in the second pinned tab, then a group tab is opened by TST automatically to put child tabs in a tree.

* Because the "Close this tab when there are no more children" is checked by default, the group tab will be closed by TST automatically when I close both 4th and 5th tabs.
* If you checked "Close this tab when there is one or less child", the group tab will be closed by TST automatically when you just close one of 4th or 5th tabs.

The option using labels defined in `config_groupTabTemporaryState_option_*` controls the default state of these checkboxes. Such group tabs are opened in more situations: for multiple new tabs opened by another extension, for multiple new tabs opened from a bookmark folder, or when a parent tab is closed but there are still some child tabs.

Does this explanation help your translation?

SammysHP:
Yes, thank you, that helps. I will update the translation as soon as possible.

SammysHP:
@piroor Updated, ready to merge. :)

piroor:
Thanks a lot!

PR #2803, opened 2021-01-26
Update messages.json RUS

wvxwxvw:
Hi! And in the 601 line of the English file, is everything correct? I just don't understand there "per 10%", it is as if superfluous in meaning. I have the impression that the artifact is left over from the previous version of the string.


piroor:
Thanks!

PR #2812, opened 2021-02-12
Add TST Select Random Tab to list of addons

54ac:
Hello,

I've created a simple addon for personal use. Maybe others will find it useful as well.

Link [here](https://addons.mozilla.org/firefox/addon/tst-select-random-tab/).


piroor:
Thanks!

BTW, it looks to work as "select random child tab" actually. I think it should be described at the page: https://addons.mozilla.org/ja/firefox/addon/tst-select-random-tab/

54ac:
Thanks for merging. It works depending on the amount of children - it looks for children first, then siblings.

Issue #2865, opened 2021-04-18
New extension: TST Tab Search


NiklasGollenstede:
Hi piroor! This isn't a bug report or feature request, but more of a thanks and FYI:

I tend to have somewhere between hundreds and thousands of tabs "open" in Firefox, and TST already immensely helps managing that. One thing I was still missing was some way of searching through and/or filtering the tabs *in the tree*. Not everything can be organized in a tree (and by far not all my tabs are organized ^^).

Since I already knew about TST's extension API, I decided to have a look into what's possible in this case, and, lo and behold, there is a `subPanel` option!

So I went ahead and implemented [TST Tab Search](https://github.com/NiklasGollenstede/tst-search). It adds a search box at the bottom of TST's sidebar, allowing to search the titles of the tabs in the current window/sidebar, optionally case sensitive, as whole word, or by regular expression. Matching tabs will be highlighted in the tree, and/or non-matches will be suppressed.
There is certainly some more potential here, but as a first version it does the job.

Primarily I wanted to give you the feedback that your API is working very well, in particular also the `subPanel` (which I think isn't used much yet) works perfectly well (for this use case, at least).
I know how tricky the communication between extensions is, because either end can show up or disappear at any point without much warning. So I am actually (positively) surprised how stable your integration API is by now.

Apart from that, before I publish this on AMO, I wanted to ask whether you had any comments or concerns, e.g. regarding the name (which may imply association with TST itself).

piroor:
Wow! You are the first challenger using the subpanel API except me! I didn't imagine it provides availability to implement this kind addon, and you done it so awesome - I'm so excited!

Now I'm inspired by your addon and planning to extend the subpanel API:

* Add ability to set fixed height for a specific addon (via API parameter)
* Add ability to memorize height of each subpanel (by TST itself)

Any other request around the API?

> Apart from that, before I publish this on AMO, I wanted to ask whether you had any comments or concerns, e.g. regarding the name (which may imply association with TST itself).

No problem, there are already some third-party helper addons including "TST" to their name :-) And, I'll post some feedbacks after I tried your addon more.

piroor:
I've implemented new parameters for `subPanel`: `initialHeight` and `fixedHeight`. For more details please see the wiki: https://github.com/piroor/treestyletab/wiki/SubPanel-API#how-to-register-a-subpanel
And now TST saves/restores the height of the subpanel area for each provider addon.

NiklasGollenstede:
[...]

Issue #2864, opened 2021-04-18
Make numbered browser tabs clickable by voice


tigersof:
[feature request] 

![image](https://user-images.githubusercontent.com/82762276/115154966-8a7d0900-a075-11eb-9579-d7d50f4ea9ee.png)

<h2>Problem Identified:</h2>
I am reliant on voice recognition and struggle to access browser tabs easily by voice.  Dragon Professional Individual has not implemented a way to achieve this in its add-on extension for Firefox and there is no Firefox ActiveX component available for Dragon’s advanced scripting.
<h3></h3>

With Dragon (see numbers on image for reference):

1. I can say “Ctrl 1” to click tab “1”, “Ctrl 2” to click tab “2”, “Ctrl 3” to click tab “3” and so on, up to “Ctrl 9” for tab “9”.
2. **Due to obvious keyboard limitations, I cannot voice-click any tabs beyond tab 9**.  I usually have a lot more tabs open, so I need the ability to voice-click those too.

My attempt at a workaround within the Dragon environment is to implement a command that performs  “Ctrl 1” and then iterates down the tabs with “Ctrl Tab” until the tab I want is reached. It works, but not always and it is inefficient and resource intensive, and often gets stuck due to pages reloading on the way down. As another example, I had also tried commands such as saying “Ctrl 123” but this merely resulted in the “Ctrl 1” outcome, and similarly for all other numbers over 9.

<h2>Feature Request:</h2>
I have been using Tree Style Tab extensively, it’s a great extension and I find it really useful.  As you can see, I have flattened the hierarchy as this better meets my needs with voice recognition, and I have inserted some of your code snippets for tab counting and numbering.

With Tree Style Tab (again, refer to numbers in image):

2. **Is there a way to make these tab numbers clickable by voice?** As one possible idea I have, can they be hyperlinked within TST, with additional perhaps hidden text (e.g. "t123" so that I can tell Dragon to click marker labelled "Tab 123" to access it) to prevent voice-clicking conflicts with other similarly named links on the page or elsewhere within the Firefox window? Or is there a better way to voice-access the tabs within TST via their number?
3. **How can I remove the column of “X”** (for closing tabs, which I can do by voice)?
4. **How can I move my tab numbers to the right hand side** (to replace the “X” column)?

I should add that I have next to no programming ability at all; most of what I do is tweaking what's already there to make it work for me. It’s really basic and usually more “miss” than “hit” but I try! 

[...]

PR #2874, opened 2021-04-26
fix API:subPanel.initial/fixedHeight

NiklasGollenstede:
Following #2865 I now tested the `subPanel.initial/fixedHeight` integration options, and the height was off.
Actually removing a bit of code fixes that. And it also makes more sense now. the `mHeight` eventually ends up being assigned directly to the `subPanel` `<iframe>`, which is what the height should be for.

### DIFF
--- a/webextensions/sidebar/subpanel.js
+++ b/webextensions/sidebar/subpanel.js
@@ -143,10 +143,9 @@ async function applyProvider(id) {
 
     mSelectorAnchor.querySelector('.label').textContent = provider.subPanel.title || provider.name || provider.id;
 
-    const headerHeight = mHeader.getBoundingClientRect().height;
     if ('fixedHeight' in provider.subPanel) {
       const unit = typeof provider.subPanel.fixedHeight == 'number' && 'px' || '';
-      mHeight = Size.calc(`${provider.subPanel.fixedHeight}${unit} + ${headerHeight}px`);
+      mHeight = Size.calc(`${provider.subPanel.fixedHeight}${unit}`);
       mHeader.classList.remove('resizable');
     }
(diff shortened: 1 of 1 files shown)


piroor:
Hmm, I intentionally added the height of the sub panel, because:

* It can be changed by some reasons - system font size, user styles, and more.
* On the other hand, the height of sub panel contents can be controlled completely by the sub panel provider author.

If TST doesn't add the height of the header, sub panel contents may be cut off accidentally. How do you think about this point?

NiklasGollenstede:
What I changed is the computation of `mHeight`. Besides being saved, restored and capped, it is (I think only) used in these assignments:
```js
const appliedHeight = Math.min(window.innerHeight * 0.66, mHeight);
mContainer.style.height = `${appliedHeight + headerSize}px`;
mSubPanel.style.height = `${appliedHeight}px`;
mTabBarContainer.style.bottom = `${appliedHeight + headerSize}px`;
```
Where `mSubPanel` is the iframe whose height I would be expecting to set with this API.
The `mContainer does indeed need the additional `headerSize`. But I think you were adding it twice before.

And I can tell you that it works now :shrug: (for me, at least)

piroor:
Oops, sorry I completely misunderstood what I did... You are right, I didn't need to add the header size!

NiklasGollenstede:
Cooli!

By the way, (and not that I am urging you) but do you have an idea when you will publish this change?

piroor:
I think #2864 is a blocker for the new release, because some changes I already did can break accessibility and need to be tested.

NiklasGollenstede:
Ok. Thanks for the info!

PR #2898, opened 2021-05-16
Restrict `subPanel.url`'s origin to the extension's bundles files

NiklasGollenstede:
It somehow feels like a bad idea to allow other extensions to (directly) show just any remote content in the `subPanel`.

This restricts the `.url` to sources in the registering extensions origin, and allows relative URLs to be interpreted in that namespace.

I am not sure how you do error reporting here. A bit more type checking and the registration or other RPC calls failing with a reported error on invalid types, would probably be a good idea in general.

Hope you don't mind the trim on the first line too much.


piroor:
Thanks a lot!

PR #2899, opened 2021-05-16
Pass the current `windowId` to the `subPanel`s

NiklasGollenstede:
I could not figure out a generic way for the `subPanel` to determine the id of the window it is in, ouside situations where it happens to be known to be the `currentWindow`.

There would be multiple possible ways for TST to pass that information. As far as I can tell, there is no direct RPC between the subpanel content and the sidebar yet. A simple `postMessage` by the sidebar would be sub-optimal, because the panel content might not be listening yet.
So I figured a URL query parameter is a simple way to transfer this information.

This information expects the `subPanel.url` to [be a valid URL](https://github.com/piroor/treestyletab/pull/2898).

### DIFF
--- a/webextensions/sidebar/subpanel.js
+++ b/webextensions/sidebar/subpanel.js
@@ -195,7 +195,7 @@ async function applyProvider(id) {
     }
 
     const url = new URL(provider.subPanel.url);
-    url.searchParams.set('tst-windowId', mTargetWindow);
+    url.searchParams.set('windowId', mTargetWindow);
     provider.subPanel.url = url.href;
 
     if (mHeight > 0)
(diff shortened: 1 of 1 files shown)


webextensions/sidebar/subpanel.js
```diff
@@ -194,6 +194,10 @@ async function applyProvider(id) {
       }
     }
 
+    const url = new URL(provider.subPanel.url);
+    url.searchParams.set('tst-windowId', mTargetWindow);
```
piroor: The prefix `tst-` looks unnecessary, I think `windowId` is enough reasonable.

piroor:
Thanks! This looks very reasonable change. Could you see my comment about the prefix?

webextensions/sidebar/subpanel.js
NiklasGollenstede: \yeah, there should be very little reason for anyone to want to have a conflicting parameter ...

piroor:
Thanks again!

NiklasGollenstede:
Ok. Done.
Thanks for the feedback and review!

PR #2900, opened 2021-05-18
Update Korean

BoredSomeone:
Add a new string translation.


webextensions/_locales/ko/messages.json
```diff
@@ -1,3 +1,4 @@
+Updated translations for 'ko': Korean (한국어):
```
piroor: This line causes parse error as a JSON file. Could you remove this?

piroor:
Sorry for a delay. I'll merge this after the parse error is corrected.

piroor:
Thanks!

PR #3264, opened 2022-12-09
README: Bookmarking trees is now a native feature

waldyrious:
I was confused by having two similar context menu entries relating to saving a tree of tabs as bookmarks. I then realized that the Bookmark Tree extension that I had installed effectively duplicates a functionality that is now available directly in TST since v3.2.0.

I added this information to the entry for the Bookmark Tree extension, to hopefully make this clearer to others.


piroor:
Thanks!

Issue #1703, opened 2018-01-01
High CPU Usage for throbber loading animation on windows with many tabs 


Lej77:
## Short description
I noticed that the throbber loading animation caused hight CPU usage when I had many tabs open. I had only 1 loading tab that used the animation and about 300 other open tabs and my CPU usage went to about 30% for an infinitely loading dummy tab that I used for testing.

## Steps to reproduce

 1. Start Firefox with clean profile.
 2. Install TST.
 3. Download  ["iperf-2.0.9-win64"](https://iperf.fr/iperf-download.php) and run the program with the command `iperf.exe -s -p 8888`.
 4. Open an infinitely loading tab by entering the following URL: [localhost:8888](url).
 5. Observe Firefox CPU usage.
 6. Stop the loading tab.
 7. Open about 100 new tabs (hold `ctrl+t`) and wait for them to finish loading.
 8. Open an infinitely loading tab by entering the following URL: [localhost:8888](url).
 9. Observe high CPU usage for Firefox.

## Expected result
Throbber loading animation' CPU usage should not vary based on number of opened tabs.

## Actual result
Throbber loading animation' CPU usage does vary based on number of opened tabs.

## Environment

 * Platform (OS): Windows 10
 * Version of Firefox: 57.0.3 also on 59
 * Version (or revision) of Tree Style Tab: 2.4.3

Lej77:
This seems to be caused by a Firefox bug. When the animated throbber element is not drawn by Firefox due to optimizations it causes High CPU usage. 

Currently this will always happen since the master throbber element used for animation synchronization is placed offscreen and Firefox will therefore not draw it. To fix this we can place it on screen and then place another element over it to hide it. (Hiding it with `opacity: 0;` or `display: none;` would not work since that would also stop Firefox from drawing it.)

Here is some CSS code that fixes this (it moves the master throbber element located in the `dummy-tabs` section to on-screen area and then covers it using the `extra-background` element):
```CSS
/* TST 2.4.3 HotFix for #1703 - Fix for Performance Issues due to Hidden Master Loading Throbber */
#dummy-tabs {
  top: 0px;
}
#extra-background {
  background: var(--tabbar-bg);
}
```

------------------------------

[...]

PR #3269, opened 2022-12-27
Fix stutters for unsynchronized loading throbbers

Lej77:
I noticed that the loading throbber animation seemed to stutter a lot when it was started. After some testing I found that this only occurred for unsynchronized throbbers. See how the bottom 5 tabs in the GIF below compares to the synchronized tabs at the top of the GIF:

loading animation

I remembered #1703 and investigated whether transparency was the problem again. Currently the background throbber is [always transparent](https://github.com/piroor/treestyletab/blob/d2ba9defe6469f785a04f127fac87d403c147607/webextensions/sidebar/styles/base.css#L271) so I added some custom CSS to disable that transparency:
```css
#dummy-tabs {
    clip: unset !important;
}
```

After that the animations no longer stuttered so that seems to fix the issue. I did some performance testing and unfortunately the change caused higher CPU usage. To minimize this disadvantage the PR makes the background throbber transparent as soon as there are no unsynchronized throbbers. Since the throbbers are only unsynchronized for a short time then maybe the higher CPU usage is worth it for the smoother animations?

# Performance

To easier observe CPU usage of loading tabs I create infinitely loadings tabs using the same method I used in issue #1703.

 2. Install TST.
 3. Open about 200 new tabs (hold `ctrl+t`) and wait for them to finish loading.
 4. Download  ["iperf-2.0.9-win64"](https://iperf.fr/iperf-download.php) and run the program with the command `iperf.exe -s -p 8888`.
 5. Open 10 infinitely loading tabs by entering the following URL: `localhost:8888`.
 6. Scroll so that tabs are not visible in native tabbar.
 7. Scroll so that the loading tabs are visibile in the sidebar.
 8. Observe Firefox CPU usage.
 9. Scroll so that the loading tabs are **not** visibile in the sidebar.
 10. Observe Firefox CPU usage.

I repeat all these steps for 3 different scenarios:

- Invisible: the background loading thobber is transparent.
  - This is the case if no custom CSS is applied to the current TST version.
- Shown: the background loading throbber is **never** transparent but is still hidden behind the sidebar background element.
  - This is the case if the following CSS is applied to the sidebar:
```css
#dummy-tabs {
    clip: unset !important;
}
```
- Dynamic: the background loading throbber is transparent if all throbbers are synchronized.
  - This is the case if this PR is applied.

## CPU usage when all throbbers are unsynchronized

To test this I disabled synchronization with the following css:

```css
/* Never sync loading throbbers */
#sync-throbber {
  animation: unset !important;
}
```

|                    | Invisible  |   Shown    | Dynamic (Same as `Shown`)  |
|--------------------|------------|------------|----------------------------|
| Scrolled into view |    4-6%    |    19%     |           19%              |
| Scroll out of view |    2%      |    14%     |           14%              |

## CPU usage when all throbbers are synchronized

|                    | Invisible  |   Shown    | Dynamic (Same as `Invisible`) |
|--------------------|------------|------------|-------------------------------|
| Scrolled into view |    6-8%    |    5-7%    |         6-8%                  |
| Scroll out of view |    2-3%    |    4-6%    |         2-3%                  |

## Environment

 * Platform (OS): Windows 10
 * Version of Firefox: 108.0.1
 * Version (or revision) of Tree Style Tab: 3.9.11

### DIFF
--- a/webextensions/common/Tab.js
+++ b/webextensions/common/Tab.js
@@ -2143,6 +2143,14 @@ Tab.getNeedToBeSynchronizedTabs = (windowId = null, options = {}) => {
   });
 };
 
+Tab.hasNeedToBeSynchronizedTab = windowId => {
+  return !!TabsStore.query({
+    windowId,
+    tabs:     TabsStore.unsynchronizedTabsInWindow.get(windowId),
+    visible:  true
+  });
+};
+
 Tab.hasLoadingTab = windowId => {
   return !!TabsStore.query({
     windowId,
--- a/webextensions/common/constants.js
+++ b/webextensions/common/constants.js
@@ -219,6 +219,7 @@ export const kTABBAR_STATE_BLOCKING               = 'blocking';
 export const kTABBAR_STATE_BLOCKING_WITH_THROBBER = 'blocking-throbber';
 export const kTABBAR_STATE_BLOCKING_WITH_SHADE    = 'blocking-shade';
 export const kTABBAR_STATE_HAVE_LOADING_TAB       = 'have-loading-tab';
+export const kTABBAR_STATE_HAVE_UNSYNCHRONIZED_THROBBER = 'have-unsynchronized-throbber';
 export const kTABBAR_STATE_THROBBER_SYNCHRONIZING = 'throbber-synchronizing';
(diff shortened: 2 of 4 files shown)


piroor:
Thanks a lot!

PR #3384, opened 2023-09-17
Improve performance of updateVisualMaxTreeLevel when animation is disabled

### LINKED ISSUE #3383: [Bug] Closing multiple tabs is slow if animation is disabled
## Abstract

When closing multiple tabs, it appears that `updateVisualMaxTreeLevel` is called once *per tab closed* if animation is disabled.

Every time a tab is closed, `reserveToUpdateVisualMaxTreeLevel` is called, which is supposed to queue/debounce calls to `updateVisualMaxTreeLevel`.

If animation is enabled, `reserveToUpdateVisualMaxTreeLevel` sets a timeout to ensure that `updateVisualMaxTreeLevel` is only called when the tabs are all finished closing. However, if animation is *disabled*, [`reserveToUpdateVisualMaxTreeLevel` will *always* call `updateVisualMaxTreeLevel`](https://github.com/piroor/treestyletab/blob/0a85f93e1caf6a9f5016c6d746ca30f6347acdaf/webextensions/sidebar/indent.js#L149-L152).

This is a problem because not only does `updateVisualMaxTreeLevel` require an expensive CSS update, but it looks through every tab in order to determine the maximum tree level. This makes performance quadratic.

For example, let's say you have 200 tabs open. That's a lot, so you decide to clean up after yourself and close 150 of them. Here's what will happen:
- Tab 1 is removed. TST calls `reserveToUpdateVisualMaxTreeLevel`, and it queries 199 tabs to figure out what the max tree level is.
- Tab 2 is removed. `reserveToUpdateVisualMaxTreeLevel` queries 198 tabs.
- Tab 3 is removed. `reserveToUpdateVisualMaxTreeLevel` queries 197 tabs.
- Tab 4 is removed. `reserveToUpdateVisualMaxTreeLevel` queries 196 tabs.
- [...]
- Tab 150 is removed. `reserveToUpdateVisualMaxTreeLevel` queries 50 tabs.

In total, we end up making 150 individual CSS updates and performing queries on *18825 tabs*. This gets much worse the more tabs you close--if you close 1000 tabs, for instance, TST will search through at least 500,500 tabs.

## Steps to reproduce

1. Start Firefox with a clean profile and open a bunch of tabs (preferably at least 100).
2. Install TST.
3. Enable [Firefox Profiler](https://profiler.firefox.com/).
4. Uncheck "Enable animation effects" in the settings panel.
5. Open the profiler dropdown menu (it's in the addon tray in the top right), select the "Firefox" preset, and start profiling.
6. Right-click the bottommost tab in the sidebar and click "Close Multiple Tabs" -> "Close Tabs to the Top".
7. Wait for TST to become responsive again, then click the profiler icon to stop profiling.
8. Observe the resulting performance profile, and note how much time is spent in `reserveToUpdateVisualMaxTreeLevel`.

[Here's a performance profile that I recorded which demonstrates the problem.](https://profiler.firefox.com/public/yh5jevnffyrjjw3sq1tx3ce8t392av2kn55cezg/stack-chart/?globalTrackOrder=01&thread=1&v=10)

## Expected result

Closing multiple tabs should be relatively fast.

## Actual result

Closing multiple tabs freezes the entire WebExtensions process for a good few minutes, and it gets worse the more tabs you have open.

## Environment

 * Platform (OS): Linux
 * Version of Firefox: 113.0.1
 * Version (or revision) of Tree Style Tab: 3.9.17

valadaptive:
Resolves #3383.

With animation disabled, we were previously updating the max tree level every time we received a tab event. When closing large numbers of tabs, this gets extremely slow.

This PR changes it to batch the updates so they only occur once per frame at most.

I wasn't sure whether to go with a 0ms delay (so it appears instantaneous) or a 100ms delay (like reserveToUpdateIndent does just below). I went with a 0ms delay in the end so as to preserve existing behavior.

Maybe for a future PR, it would be nice to do something like what [Lodash.debounce](https://docs-lodash.com/v4/debounce/) offers with `options.leading`: the *first* time we receive the event, we apply it instaneously, but every event after that gets queued up with a certain delay. This would prevent visual feedback from being too delayed, while giving a good "buffer zone" to avoid updating too frequently.

### DIFF
--- a/webextensions/sidebar/indent.js
+++ b/webextensions/sidebar/indent.js
@@ -146,15 +146,12 @@ export async function reserveToUpdateVisualMaxTreeLevel() {
     delete updateVisualMaxTreeLevel.waiting;
   }
 
-  if (!shouldApplyAnimation()) {
-    updateVisualMaxTreeLevel();
-    return;
-  }
+  const delay = shouldApplyAnimation() ? configs.collapseDuration * 1.5 : 0;
 
   updateVisualMaxTreeLevel.waiting = setTimeout(() => {
     delete updateVisualMaxTreeLevel.waiting;
     updateVisualMaxTreeLevel();
-  }, configs.collapseDuration * 1.5);
+  }, delay);
 }
 
 function updateVisualMaxTreeLevel() {


piroor:
Thanks a lot, but I have a concern: this change introduces a visual regression. When a deeply nested tree is expanded, you'll see an expanded tree with old indentations and it is updated to correct indentations with a delay. Hmm, how can we avoid that...

piroor:
I close this PR because it is outdated by the change 5fa5f8d which is inspired from this PR. Thanks again!

PR #3446, opened 2024-01-31
Add "bookmarks" to manifest.json as it now required in nightly builds

adderek:
Recent update of firefox nightly broke the extension. Currently permission "bookmarks" is required.


piroor:
I close this because this change is needless on TST 3.9.22 and later.

PR #3753, opened 2025-06-13
Fix contextTab error on sidebar right-click on free space

joanbm:
Due to the recent changes by commit 973555f0b147 ("Invoke commands in fake context menu correctly"), right-clicking on free sidebar space then selecting an option of the context menu (e.g. "Reopen Closed Tab") causes a 'contextTab is null' error, since the context menu wasn't opened over any tab, but the changes assume `contextTab` isn't null.

tstbug25

Fix it by tolerating a null `contextTab`.


piroor:
Thanks!

PR #3763, opened 2025-06-24
Updated translations for 'uk': Ukrainian (Українська):

Taron-art:
Hi.

I've tried to partially update the translation using WET. It looks different (contains hashes). Is it OK? It is hard to compare changes. Hope this works, thanks.


piroor:
Thanks a lot!

PR #3769, opened 2025-07-05
Update Ukrainian translation

Taron-art:
Hello.

Thanks for approving the previous PR, I've created some more updates. Plan to continue may be next weekend. Sorry for a non-complete PRs, I cannot properly save and continue on WET (or, most probably, do not know how :) ).


piroor:
Thanks again!

PR #3770, opened 2025-07-06
Fix container assignment for extra tab contents below tabs

emmanuel-ferdman:
## PR Summary
This small PR fixes a bug routing 'tab-above' content to below-container instead of above-container.


piroor:
Thanks!

PR #3832, opened 2026-02-09
Fix bugs in tree-structure restoration and progress tracking

tkng:
Fix three bugs in `webextensions/background/tree-structure.js` related to tab restoration and progress tracking.

### 1. "Undo Close Tabs" never triggered for multiple tabs

`Array.prototype.filter()` returns an array, but the result was compared directly as `countToBeRestored > 0`. Since comparing an array to a number always evaluates to `false`, `tryRestoreClosedSetFor()` was never called. Added the missing `.length` property access.

### 2. Wrong progress calculation during tab restoration

`Math.round(maxCount - count / maxCount * 100)` was evaluated as `maxCount - ((count / maxCount) * 100)` due to operator precedence, producing incorrect progress values. Fixed to `(maxCount - count) / maxCount * 100`.

### 3. Dead condition in Tab.onRemoved listener

`mRecentlyClosedTabs` was cleared to `[]` and then immediately checked with `.length > 1`, so `newlyRestorable` was always `false`. The condition `currentlyRestorable != newlyRestorable` only passed when `currentlyRestorable` was `true`, dispatching `false` — which was the correct value but via unnecessarily convoluted logic. Simplified to directly check the pre-clear state.

### DIFF
--- a/webextensions/background/tree-structure.js
+++ b/webextensions/background/tree-structure.js
@@ -452,7 +452,7 @@ Tab.onRestored.addListener(tab => {
             UserOperationBlocker.unblockIn(tab.windowId, { throbber: true });
           }, 0);
 
-          const countToBeRestored = mRecentlyClosedTabs.filter(tab => !mRestoredTabIds.has(tab.uniqueId));
+          const countToBeRestored = mRecentlyClosedTabs.filter(tab => !mRestoredTabIds.has(tab.uniqueId)).length;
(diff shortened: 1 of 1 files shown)


piroor:
Thanks!

BTW, please send PRs for each topic at the next chance, instead of a single PR containing multiple different topics, if possible. I maintain the changelog based on summaries of git log, so I accidentally miss detailed topics from multi-topics commits.

Issue #3881, opened 2026-03-05
[Bug] Tab Label text-alignment/direction and overflow behavior doesn't match native tab bar


tcbbd:
## The overflow issue

When the tab label is too long, it fades out. Simple and elegant. But what if the tab label is in arabic? It turns out we must consider the text direction of tab label.

TST handles the overflow quite well if the browser language is LTR. But if we change the browser language to RTL, we get this:

<img width="2844" height="1775" alt="Image" src="https://github.com/user-attachments/assets/6cbece3d-873c-4fe9-b283-a7c31fdd4751" />

Notice that the native tab bar recognizes the text direction of tab label, and let the label fade naturally. But TST cannot do that. The fading effect changes according to text direction, but the text itself doesn't. It seems we need to set `direction: ltr` for overflowing LTR labels to match the native tab bar.

## The general text alignment/direction issue

This is an off topic from #3869. In that post, I said that as we plan to change the position/alignment and possibly text direction of collapsed members counter, we may also want to consider the same thing for all other tab labels. After some investigation, I found that the native tab bar behaves this way:

- If the browser language is LTR
  - RTL tab labels are set to rtl text direction, so the text aligns to right
- If the browser language is RTL
  - LTR tab labels are not set to ltr text direction,  so the text aligns to right
  - unless the tab label overflows, in which case it is set to ltr text direction, so that the fading looks natural.

When the browser language is LTR, TST only sets RTL tab labels to rtl text direction when they overflow. This doesn't match native tab bar. And when the browser language is RTL, we should set LTR tab group labels to ltr text direction, otherwise it looks ugly as shown in the screenshot.

---

In #3869, I got the idea that the tab label alignment can be designed to change according to sidebar side. A possible design could be like:

- When sidebar style is left side, align all tab labels to left, whether it's LTR or RTL
- When sidebar style is right side, align all tab labels to right, whether it's RTL or LTR
- If the tab label overflows, set the text direction according to its content

But clearly, this does not match native tab bar behavior. If we want to match native tab bar in some sense, it could be like this:

- When sidebar style is left side, we treat the tab labels as if the browser language is LTR
  - LTR labels align to left, RTL labels align to right
- When sidebar style is right side, we treat the tab labels as if the browser language is RTL
  - LTR and RTL labels all align to right

[...]

Issue #3869, opened 2026-02-27
[Bug] Tab groups are interwoven when partially collapsed (Edit: plus other tab group issues)


tcbbd:
## Steps to reproduce

1. Create 2 tab groups, say group A and group B.
2. Open some tabs in each group.
3. Select a tab in group A, and fold group A.

## Expected result

Group B follows Group A.

## Actual result

Group A and Group B are interwoven:

https://github.com/user-attachments/assets/d7da29bb-0dcb-45b7-9f62-12f2798a27e4

## Environment

 * Platform (OS): Windows 11
 * Version of Firefox: 148.0 Stable
 * Version (or revision) of Tree Style Tab: 4.2.9

tcbbd:
Btw, it seems that the right side layout is for RTL scripts. But tab group labels are left aligned instead of right aligned. If I understand correctly, they should be right aligned for an RTL layout.

<img width="512" height="573" alt="Image" src="https://github.com/user-attachments/assets/55c91b36-40b2-4907-bb5a-8e656825f269" />

tcbbd:
Another bug:

https://github.com/user-attachments/assets/18d65180-276f-4489-a46d-853e72043ebd

As you can see, the group always shows "+5" as  the number of hidden tabs. It cannot reflect the change of shown/hidden tabs. I think this is not desired and should be considered a bug.

piroor:
Thank you for reporting! The bug about wrong position of the collapsed members counter has been fixed with recent commits.

However I've not reproduced the wrong alignment issue, with Firefox ESR140(ar) + TST latest development build. Screenshot:
<img width="1049" height="827" alt="Image" src="https://github.com/user-attachments/assets/e86e0325-126e-4b1a-8c5b-1bf018e30255" />
Could you try the latest development build?: https://github.com/piroor/treestyletab?tab=readme-ov-file#development-builds

tcbbd:
Interesting. It works if I change the browser language to arabic.

It seems that for RTL scripts, its logic is: Tab group label always align to right + Sidebar style always right side. For LTR scripts, its logic is: Tab group label always align to left + Sidebar style depends on the side (left for left side, right for right side). If you choose left side sidebar style manually for RTL scripts, the alignments of tab groups and tabs will mismatch.

[...]

PR #3888, opened 2026-03-17
Various stylesheet fixes (#3869 #3881)

tcbbd:
The stylesheet tweaks for #3869 and #3881 as requested, plus some more fixes.

### DIFF
--- a/webextensions/sidebar/styles/photon/base.css
+++ b/webextensions/sidebar/styles/photon/base.css
@@ -76,7 +76,8 @@ tab-item.pinned tab-item-substance {
 :root.regular .after-tabs [role="button"]:not([data-menu-ui]) {
   padding-inline-end: 0.35em;
 }
-:root.regular tab-item:not(.pinned)[data-parent-id] tab-item-substance {
+:root.regular tab-item:not(.pinned)[data-parent-id] tab-item-substance,
+:root.regular tab-item:not(.pinned)[data-group-id]:not([data-group-id="-1"]) tab-item-substance {
   border-inline-start-width: 1px;
 }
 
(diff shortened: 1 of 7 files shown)


tcbbd:
And I think it looks better if we remove this `padding: 0` for Classic Sidebar theme:

```css
tab-item[type="group"] {
  tab-item-substance,
  .ui {
    padding: 0 !important;
  }
```

Because if we have round corners, we should also add some padding space. Zero padding space works better with square corners. But I'm not sure whether it is intentionally set to 0, so I didn't touch this line.

webextensions/sidebar/styles/photon/base.css
```diff
@@ -32,6 +32,10 @@ tab-item[type="group"] {
     padding: 0 !important;
   }
 
+  tab-item-substance {
+    z-index: 110;
```
piroor: This should be a relative value like `calc(var(--tab-base-z-index) + 10)` if it aims to be larger than the z-index of regular tabs always.

tcbbd: Fixed.

piroor:
> And I think it looks better if we remove this `padding: 0` for Classic Sidebar theme:
> 
> ```css
> tab-item[type="group"] {
>   tab-item-substance,
>   .ui {
>     padding: 0 !important;
>   }
> ```
> 
> Because if we have round corners, we should also add some padding space. Zero padding space works better with square corners. But I'm not sure whether it is intentionally set to 0, so I didn't touch this line.

I'm OK to remove this. I ordinary use TST + Firefox on Windows with Proton, so sometimes nonsense rules may be left in Sidebar and Photon...

tcbbd:
> I'm OK to remove this. I ordinary use TST + Firefox on Windows with Proton, so sometimes nonsense rules may be left in Sidebar and Photon...

I use Photon daily. Just to ensure the changes don't break any theme, I checked all the themes.

piroor:
Thanks a lot!

SteveOnorato:
Nothing critical, I have a workaround, but I just wanted to flag that I think this caused a minor regression for my configuration: https://github.com/piroor/treestyletab/discussions/3905

PR #3934, opened 2026-06-12
Add remove-tab-keeping-children API for helper addons

### LINKED ISSUE #3933: API request: close a parent tab while preserving children (without requiring visual expansion)
Hello, I'm the developer of [Duplicate Tabs Closer (DTC)](https://github.com/Peuj/duplicate-tabs-closer), an extension that automatically detects and closes duplicate browser tabs.

## The problem

When DTC closes a duplicate tab that is also a TST parent with collapsed children, TST cascade-closes all descendants together with the parent. This is TST's correct default behavior, a collapsed parent closing takes its children with it, but not expected by the user.

DTC's current workaround is a two-step sequence:
1. Call `{ type: "expand-tree", tab: tabId }` to expand the collapsed parent
2. Close the parent tab via `browser.tabs.remove()`

After step 2, TST sees an expanded parent being removed and correctly auto-promotes the first child as the new root parent of the remaining siblings. The children survive.

## The limitation of the workaround

This approach has three drawbacks:

1. **Two IPC round-trips** (`get-tree` to detect collapsed children, then `expand-tree`) instead of one
2. **Visual flash**: children briefly appear visible in the TST sidebar before the parent disappears, which is noticeable to the user
3. **Timing dependency**: the close fires immediately after the `expand-tree` IPC resolves; any internal TST async processing between those two steps is a potential race

## The request

A new API message (or a flag on an existing command) that tells TST: "Close this tab, but apply the expanded-parent behavior when handling its children, auto-promote them instead of cascade-closing them, regardless of their current collapse state."

Something like:
```js
await browser.runtime.sendMessage("treestyletab@piro.sakura.ne.jp", {
  type: "remove",
  tab: tabId,
  keepChildren: true   // or: closeDescendants: false
});
```

This would let DTC express its intent in a single atomic call, without needing to orchestrate TST's visual state first.

## Current implementation

DTC already ships the `expand-tree` workaround as of version 4.2.1. The relevant code is in [`tst.js`](https://github.com/Peuj/duplicate-tabs-closer/blob/master/tst.js). Happy to update it to use a cleaner API if one becomes available.

BTW, thank you for TST, it's a mandatory extension for me since a long time now!

Peuj:
Closes #3933

Adds a new external addon API message type `remove-tab-keeping-children`
that allows helper addons to close a parent tab while promoting all its
children, without the visual expand/collapse flash of the current
expand-tree + remove two-step approach.

## How it works

The handler (in `onMessageExternal`):
1. Finds children of the target tab via native `openerTabId`
2. Updates each child's `openerTabId` to the grandparent (or self if tab
   is a root) before removing the parent
3. Calls `browser.tabs.remove()` on the parent

This is pure Firefox API with no dependency on TST's internal tree
tracking state, so it works regardless of whether TST auto-attached
the children.

## Usage

```js
browser.runtime.sendMessage('treestyletab@piro.sakura.ne.jp', {
  type: 'remove-tab-keeping-children',
  tab:  tabId,          // or tabs: [tabId1, tabId2]
});
```

Returns `true` on success.

## Tests

Two new tests added in `test-close-parent-behavior.js`:
- `testRemoveTabKeepingChildrenWithCollapsedSubtree`
- `testRemoveTabKeepingChildrenWithExpandedSubtree`

Both pass. No regressions vs clean trunk.


webextensions/background/handle-misc.js
```diff
@@ -1,1167 +1,1188 @@
-/*
-# This Source Code Form is subject to the terms of the Mozilla Public
-# License, v. 2.0. If a copy of the MPL was not distributed with this
-# file, You can obtain one at http://mozilla.org/MPL/2.0/.
-*/
+/*
```
piroor: Too much diff - it looks to be caused by changed line-endings (LF => CR+LF) happened by mixed line-endings unexpectedly introduced by me.  Could you uniform line-endings with LF?

Peuj:
Sorry about that. core.autocrlf=true on my machine caused the whole file to be converted to CRLF when I edited it. Should be fixed now.

webextensions/background/handle-misc.js
```diff
@@ -1109,6 +1109,27 @@ function onMessageExternal(message, sender) {
...
 
+    case TSTAPI.kREMOVE_TAB_KEEPING_CHILDREN:
+      return (async () => {
+        const tabs = await TSTAPI.getTargetTabs(message, sender);
+        for (const tab of tabs) {
+          if (!TabsStore.ensureLivingItem(tab))
+            continue;
+          // Re-parent children via native openerTabId BEFORE removing, so treeStructure() sees correct state.
+          // TST uses openerTabId == tab.id as a self-reference to signal "no opener", treat that as null.
+          const allWindowTabs = await browser.tabs.query({ windowId: tab.windowId });
```
piroor: I think you should detach children of the tab with `Tree.detachAllChildren()`, like as:

```javascript
const behavior = TreeBehavior.getParentTabOperationBehavior(tab, {
  context: Constants.kPARENT_TAB_OPERATION_CONTEXT_CLOSE,
}) :
await Tree.detachAllChildren(tab, {
  broadcast: true,
  behavior,
});
```

`TreeBehavior.getParentTabOperationBehavior()` returns a behavior code based on user preferences, and `Tree.detachAllChildren()` operates tree gracefully, for example it promotes the first child tab as the new parent.

Peuj:
Good point, using TST's own tree operations is much cleaner. Updated to use Tree.detachAllChildren with PROMOTE_ALL_CHILDREN behavior, which handles the collapsed subtree case automatically too.

Peuj:
After digging into the test setup, I found that the test creates tabs with openerTabId but autoAttachOnOpenedFromExternal=kNEWTAB_DO_NOTHING, so TST's $TST.children is empty and Tree.detachAllChildren has nothing to operate on. The tree structure in the tests is measured via native openerTabId (treeStructure() checks tab.openerTabId first), so updating openerTabId directly is the correct approach here.

If you would like both approaches (native openerTabId for compatibility + Tree.detachAllChildren when $TST children are populated), happy to add that too.

piroor:
I've introduced some more changes to refine the API with the PR #3952 which alternates this PR, thus I close this. Thank you for the base implementation!

PR #3941, opened 2026-07-26
Fix typos in en locale

scrypt-kitty:
Six user-visible English strings in `webextensions/_locales/en/messages.json` contain misspellings or subject–verb agreement errors. This PR corrects only those, with no other changes.

| Key | Before | After |
|---|---|---|
| `sentTabs_notification_message` | "**Tabs is** sent to $DEVICE$" | "Tab is sent to $DEVICE$" |
| `config_syncRequirements_description` | "activate Firefox Sync with **ther** browser's options" | "with the browser's options" |
| `config_requireBookmarksPermission_tooltiptext` | "request **an required** permission" | "a required permission" |
| `config_autoDiscardTabForUnexpectedFocus_label` | "when tabs are **accidentaly** focused" | "accidentally focused" |
| `config_tabDragBehavior_noteForDragstartOutsideSidebar` | "When a **parent tabs is** dragged" | "When a parent tab is dragged" |
| `config_externaladdonpermissions_description` | "even if the extension **ifself** is forbidden" | "the extension itself" |

### On `sentTabs_notification_message`

This one is singular, not plural. `common/sync.js` picks the message key by appending a `_multiple` suffix, and the sibling `sentTabs_notification_message_multiple` already reads "Tabs **are** sent to $DEVICE$" — so this slot is the one-tab case and the fix is "Tab is sent", not "Tabs are sent".

### Misspelled keys left alone deliberately

Some option *keys* are also misspelled, but renaming them would break existing installs, so they are untouched here:

- `sidebarPositionRighsideNotificationShown` — a persisted config, also referenced by `background/migration.js`
- `autoExpandOnLongHoverRestoreIniitalState` — a persisted config
- `tabbar_newTabWithContexualIdentity_*` — not a config, but renaming it would orphan the key in the eight other locales

### Verification

- The file parses as JSON and the key count is unchanged (813 before and after).
- The key set is byte-identical to `trunk`, so no translation in any other locale is orphaned. No other locale files are touched.
- Only these six `message` values differ from `trunk`.

Found while reviewing Waterfox's vendored copy of TST (BrowserWorks/waterfox#4324). Two other typos from that review — `conetnt` and `broat` — are already fixed here in eb2ff88d0 and fdfc45a9e; that fork is simply carrying an older snapshot.

### DIFF
--- a/webextensions/_locales/en/messages.json
+++ b/webextensions/_locales/en/messages.json
@@ -224,7 +224,7 @@
   "syncAvailable_notification_message": { "message": "Tabs can be transferred between your devices. Please click here to set an identifiable name to this device." },
   "syncAvailable_notification_message_linux": { "message": "Tabs can be transferred between your devices. Please click the button to set an identifiable name to this device." },
   "sentTabs_notification_title":            { "message": "\u200b" },
-  "sentTabs_notification_message":          { "message": "Tabs is sent to $DEVICE$",
+  "sentTabs_notification_message":          { "message": "Tab is sent to $DEVICE$",
     "placeholders": {
       "DEVICE": { "content": "$1", "example": "Firefox on Device X" }
     }},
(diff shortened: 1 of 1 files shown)


piroor:
Thanks!

piroor:
>     * `sidebarPositionRighsideNotificationShown` — a persisted config, also referenced by `background/migration.js`

This is left just for migration, so we need to keep it with typo.

>     * `autoExpandOnLongHoverRestoreIniitalState` — a persisted config

This typo has been fixed with the commit dc9f2b5.

>     * `tabbar_newTabWithContexualIdentity_*` — not a config, but renaming it would orphan the key in the eight other locales

This typo has been fixed with the commit 6ad49a6.

PR #3942, opened 2026-07-26
Improve English phrasing in en locale

scrypt-kitty:
Follow-up to #3941. That PR fixes outright typos; this one is purely about English phrasing, so it is separate and entirely optional — please feel free to close it if you prefer the current wording.

| Key | Before | After |
|---|---|---|
| `sidebarPositionInvertedNotification_message` | "the one **matching to** the current position" | "the one matching the current position" |
| `message_startup_description_sync_before` | "activate Firefox Sync **with** Firefox's options **at first**, and install TST **to** other devices **also**" | "activate Firefox Sync in Firefox's options first, and install TST on your other devices as well" |
| `config_syncRequirements_description` | same phrasing as above, with "the browser's options" | same correction as above |
| `config_requireBookmarksPermission_tooltiptext` | "This feature needs **a right to** access bookmarks. Click here to request **a** required permission." | "This feature needs permission to access bookmarks. Click here to request the required permission." |

Notes on the individual changes:

- **"matching to"** — "matching" takes a direct object in English; the preposition is not used here.
- **"at first"** — this means "initially, but not later" (as in "at first I thought…"). For "before anything else", English uses "first".
- **"install TST to other devices"** — software is installed *on* a device.
- **"needs a right to access"** — grammatical, but "right" is a legal term in English. For browser permissions the idiomatic word is "permission", which also matches the wording used elsewhere in the locale.

### Verification

- The file parses as JSON and the key count is unchanged (813).
- The key set is identical to `trunk`, so no translation is orphaned. No other locale files are touched.
- Only these four `message` values differ from #3941.

Since GitHub cannot base a cross-fork PR on another fork's branch, this PR is stacked on #3941 and its diff currently shows both commits. Once #3941 is merged this will reduce to the four changes above on its own — happy to rebase at any point, or to squash both into a single PR if you would rather review them together.

### DIFF
--- a/webextensions/_locales/en/messages.json
+++ b/webextensions/_locales/en/messages.json
@@ -224,7 +224,7 @@
   "syncAvailable_notification_message": { "message": "Tabs can be transferred between your devices. Please click here to set an identifiable name to this device." },
   "syncAvailable_notification_message_linux": { "message": "Tabs can be transferred between your devices. Please click the button to set an identifiable name to this device." },
   "sentTabs_notification_title":            { "message": "\u200b" },
-  "sentTabs_notification_message":          { "message": "Tabs is sent to $DEVICE$",
+  "sentTabs_notification_message":          { "message": "Tab is sent to $DEVICE$",
     "placeholders": {
       "DEVICE": { "content": "$1", "example": "Firefox on Device X" }
     }},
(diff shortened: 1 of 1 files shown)


piroor:
Thanks!

PR #3952, opened 2026-08-25
Refined: Add remove-tab-keeping-children API for helper addons #3934

piroor:
This refines the new API introduced by #3934 with stability and flexibility.


piroor:
Document: https://github.com/piroor/treestyletab/wiki/API-for-other-addons#close-only-parent-tabs-keeping-their-children

PR #3972, opened 2026-09-19
Fix color handling for Nova themes + add theme opt-out to fix dark mode

Robertof:
Big fan of TST, thanks for your hard work! Got a nice "surprise" when updating to FF 157 on the aurora channel.

Firefox 157's new UI (Nova) introduces a bunch of new "default" themes (which I suspect a lot of people will flock to given the default palette). Unlike previous themes, those express colors as RGB arrays (e.g. `[255, 255, 255]`) rather than CSS strings. They also use linear gradients as their backgrounds. TST does not handle either properly which results in the TST stylesheet becoming borked.

In addition, a longstanding bug (https://bugzilla.mozilla.org/show_bug.cgi?id=1542044) prevents TST from getting the dark mode color palette for themes which support both dark and light mode. Because this would result in the sidebar being rendered with light mode palettes while the browser is in dark mode (and there's no way to get the dark mode palette), add a flag to allow users to opt-out of theme color palettes, thus reverting to the default TST dark mode colors which are perfectly fine.

| Scenario | |
| - | - |
| Before, Dusk theme on FF 157 | image |
| After, Dusk theme on FF 157 (flag disabled, default) | image |
| After, Dusk theme on FF 157 (flag enabled) | image |


piroor:
Thanks!

